PR #22, opened 2020-11-05
fix: manifest and setup.py fixed. close #20

### LINKED ISSUE #20: Update metadata in pypi
Currently the metadata that we have for `faust-straming` in [pypi](https://pypi.org/project/faust-streaming/) is pointing to the original `faust`. We need to update the `setup.py` and probably the `MANIFEST.in`

marcosschroh:


faust/__init__.py
```diff
@@ -25,8 +25,8 @@
 
 __version__ = "0.1.0"
 __author__ = "Robinhood Markets, Inc."
-__contact__ = "contact@fauststream.com"
-__homepage__ = "http://faust.readthedocs.io/"
+__contact__ = "schrohm@gmail.com"
```
patkivikram, line 28: you can add mine too vpatki@wayfair.com

patkivikram (approved):

faust/__init__.py
marcosschroh: yes!! sorry

PR #30, opened 2020-11-13
Restore testing

forsberg:
A lot of test classes are named test_*, which makes pytest *not* find them by default

## Description

After the change in how tests are run, only 440 tests are found, instead of 1973 tests found when robinhood/faust runs their test suite.

Main reason for this is that a lot of the test classes are named `test_*` and therefore are not automatically discovered by pytest.

This PR restores discovery. Unfortunately a bunch of tests fails on the current codebase..


forsberg:
@patkivikram @marcosschroh - I assume the non-discovery of 1533 of the tests was by accident, not on purpose? 

I don't quite feel I know enough about aiokafka to fix all those failures. Some of them seem related to transaction support, which was something that was recently remade, right? 

So, eh, I'm leaving this one here as an unwelcome gift right before the weekend. Sorry! :-)

max-k:
We have to fix it urgently because it's a huge regression regarding original project.
I'll try to help.

taybin:
I've updated @forsberg's branch with the latest master and created a new PR #141 for it.

taybin:
Because #141 was merged in, this can be closed.

PR #36, opened 2020-11-16
Return float from Cython implementation of HoppingWindow.ranges

forsberg:
## Description

This PR addresses robinhood/faust#674, where using the Cython implementation of HoppingWindow makes the table functionality (at least together with RocksDB) worthless.

It also adds one dimension to the matrix in the github action, making the tests be run with and without Cython enabled, trying to catch any future bugs like this. 

The test included in the PR has been proven to fail on Cython without the patch to the actual code, see [this test run](https://github.com/forsberg/faust-streaming/actions/runs/365487074)

As the Cython code is not counted in coverage, I had to decrease the accepted coverage percenage to 60%, or the coverage enforcement would always fail on Cython since windows.py was not counted as covered. There may be better ways to do solve this.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,12 +11,18 @@ on:
 
 jobs:
   tests:
-    name: "Python ${{ matrix.python-version }}"
+    name: "Python ${{ matrix.python-version }}/Cython: ${{ matrix.use-cython }}"
     runs-on: "ubuntu-latest"
 
     strategy:
+      # Complete all jobs even if one fails, allows us to see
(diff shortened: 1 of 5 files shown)


patkivikram (approved):

patkivikram:
Can you check why the other pipelines are not working? Once they succeed we can merge this one

forsberg:
@patkivikram - you mean the "Python 3.6/3.7/3.8" which are in "Expected - Waiting for status to be reported"?

I suspect these are github CI being confused about me replacing one set of jobs with another, so I think they should be disregarded.

PR #40, opened 2020-11-18
recovery: _slurp should only stop recovery if fetching already started

trauter:
this fixes a bug when getting highwaters/offsets takes longer than 5s.
In this case, the in_recovery flag would have been resetted because of

## Description
copy of https://github.com/robinhood/faust/pull/573

that version series.

### DIFF
--- a/faust/tables/recovery.py
+++ b/faust/tables/recovery.py
@@ -112,6 +112,7 @@ class Recovery(Service):
 
     completed: Event
     in_recovery: bool = False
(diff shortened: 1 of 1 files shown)


trauter:
since there have been changes in the recovery-logic, this PR may be obsolete now.

patkivikram:
If this is not required can you close this PR?

trauter:
Chances are high that this issue is already resolved.

PR #41, opened 2020-11-18
Change RocksDB to check the assigned topic name

g-clef:
Changes the RocksDB store to check the assigned topic name for a changelog, rather than the auto-generated one. If the changelog uses a manually-created topic, using the auto-generated name causes keys(), values(), items() on the Table object that this store backs to return no data, since the changelog names never match.

### DIFF
--- a/faust/stores/rocksdb.py
+++ b/faust/stores/rocksdb.py
@@ -400,7 +400,7 @@ class Store(base.SerializedStore):
 
     def _dbs_for_actives(self) -> Iterator[DB]:
         actives = self.app.assignor.assigned_actives()
-        topic = self.table._changelog_topic_name()
+        topic = self.table.changelog_topic_name
         for partition, db in self._dbs.items():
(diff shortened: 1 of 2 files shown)


faust/stores/rocksdb.py
```diff
@@ -400,7 +400,7 @@ def _dbs_for_key(self, key: bytes) -> Iterable[DB]:
 
     def _dbs_for_actives(self) -> Iterator[DB]:
         actives = self.app.assignor.assigned_actives()
-        topic = self.table._changelog_topic_name()
+        topic = self.table.changelog_topic_name
```
patkivikram, line 403: Isn't this a method name? I don't see a property with that name

g-clef: It's defined as a property at https://github.com/faust-streaming/faust/blob/be1e6dbe2507fd068a40391dcd2ddd436372661e/faust/tables/base.py#L608

patkivikram (approved):

Issue #53, opened 2020-12-04
Prevent stream buffer overflow by lowering the rate of incoming partitions


patkivikram:
## Checklist

- [ x] I have included information about relevant versions
- [x ] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Have an agent process slowly. When the number of events on its Queue > `stream_buffer_maxsize` it causes starvation of other stream processors(agents) processing other topics
## Expected behavior

Stop fetching data from the slow topic/partition when the buffer reaches a threshold and then resume it after the pressure drops

## Full traceback

```pytb
Paste the full traceback (if there is any)
```

# Versions

* Python version 3.8
* Faust version 0.3.0
* Operating system centos 7.5
* Kafka version 2.4
* RocksDB version (if applicable) 6.10.3

PR #55, opened 2020-12-07
Add backpressure to slow stream processing to avoid filing up the buffer

patkivikram:
This PR fixes [53](https://github.com/faust-streaming/faust/issues/53)

### DIFF
--- a/faust/transport/_cython/conductor.pyx
+++ b/faust/transport/_cython/conductor.pyx
@@ -17,6 +17,9 @@ cdef class ConductorHandler:
         object acquire_flow_control
         object consumer
         object wait_until_producer_ebb
+        object consumer_on_buffer_full
+        object consumer_on_buffer_drop
+
 
     def __init__(self, object conductor, object tp, object channels):
         self.conductor = conductor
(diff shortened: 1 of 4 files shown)


faust/transport/_cython/conductor.pyx
```diff
@@ -78,6 +96,18 @@ cdef class ConductorHandler:
         await chan.put(event)
         delivered.add(chan)
 
+    # callback called when the queue is under high pressure/
+    # about to become full.
+    def on_pressure_high(self) -> None:
```
forsberg, line 101: I'm not so familiar with Cython, but would this be faster if it was declared a cpdef? Can it be a cdef? I guess not, as it's passed to another function as an argument.

faust/transport/_cython/conductor.pyx
```diff
@@ -78,6 +96,18 @@ cdef class ConductorHandler:
...
+        self.on_topic_buffer_full(self.tp)
+        self.consumer_on_buffer_full(self.tp)
+
+    # callback used when pressure drops.
+    # added to Queue._pending_pressure_drop_callbacks
+    # when the buffer is under high pressure/full.
+    def on_pressure_drop(self) -> None:
```
forsberg, line 108: Same as above (cpdef/cdef).

forsberg (commented):
How about unit tests?

austinnichols101 (approved):
Fast.

Issue #54, opened 2020-12-06
IllegalStateException on seek to offset of a partition that was removed by a rebalance


bobh66:
## Checklist

- [X] I have included information about relevant versions
- [ ] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

faust-streaming 0.3.0
aiokafka 0.7.0

In this example, partitions 1, 5 and 10 were originally assigned but after the rebalance partition 1 is no longer assigned.  The code tries to seek to the committed offset for partition 1 and crashes:
```
[2020-12-06 15:31:12,106] [11] [WARNING] Heartbeat failed for group spe-5 because it is rebalancing 
[2020-12-06 15:31:12,106] [11] [INFO] Revoking previously assigned partitions 
+Topic Partition Set----------+
| topic          | partitions |
+----------------+------------+
| spe_requests_5 | {1, 5, 10} |
+----------------+------------+ for group spe-5 
[2020-12-06 15:31:12,108] [11] [INFO] (Re-)joining group spe-5 
[2020-12-06 15:31:12,108] [11] [INFO] [^---Recovery]: Resuming flow... 
[2020-12-06 15:31:12,109] [11] [INFO] [^---Recovery]: Seek stream partitions to committed offsets. 
[2020-12-06 15:31:12,121] [11] [INFO] Joined group 'spe-5' (generation 968) with member_id faust-0.3.0-9126a240-a759-4e9f-bca9-cc0ad3f20f7e 
[2020-12-06 15:31:12,128] [11] [INFO] Successfully synced group spe-5 with generation 968 
[2020-12-06 15:31:12,129] [11] [INFO] Setting newly assigned partitions 
+Topic Partition Set----------+
| topic          | partitions |
+----------------+------------+
| spe_requests_5 | {5, 10}    |
+----------------+------------+ for group spe-5 
[2020-12-06 15:31:12,131] [11] [ERROR] [^---Recovery]: Crashed reason=IllegalStateError("No current assignment for partition TopicPartition(topic='spe_requests_5', partition=1)") 
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/mode/services.py", line 802, in _execute_task
    await task
  File "/usr/local/lib/python3.7/site-packages/faust/tables/recovery.py", line 349, in _restart_recovery
    await T(self._resume_streams)()
  File "/usr/local/lib/python3.7/site-packages/faust/tables/recovery.py", line 297, in _resume_streams
[...]

PR #57, opened 2020-12-07
Fix for rebalance Exception causing recovery to crash

patkivikram:
Fix for [54](https://github.com/faust-streaming/faust/issues/54)


austinnichols101 (approved):
Works great - thanks @patkivikram !

PR #66, opened 2020-12-21
Fix IPv6 address case in server_list function

Anvil:
## Description

When specifying an IPv6 address in faust.App as broker, faust passes to aiokafka an incorrectly-transformed value. 

e.g:

When I write this:
```python
faust.App('mob', broker=f'kafka://[1234:5678::9abc]')
```

yarl.URL constructor receives the strings as is, but the returned object `.host` attribute is `1234:5678::9abc`.  So, the `server_list function` in transport/drivers/aiokafka.py creates `1234:5678::9abc:9092`, which, while not syntactically incorrect, is a different IPv6 address. Also, the port information is lost in the process.

I've checked the yarl module and there does not seem to be any attribute/method to retrieve the enclosed IPv6 address. This small change intend to fix this behavior, on faust side.

L156 is assumed as a "is it an IPv6 address" check. While it could be better, I found it's good enough for me. :)

### DIFF
--- a/faust/transport/drivers/aiokafka.py
+++ b/faust/transport/drivers/aiokafka.py
@@ -159,10 +159,24 @@ Has not committed %r (last commit %s).
 """.strip()
 
 
+def __canon_host(host, default):
+    """Ensure host is correctly formatted for aiokafka. That means IPv6
+    addresses must enclosed in squared brackets.
+    """
+    if not host:
+        return default
+    if ":" in host:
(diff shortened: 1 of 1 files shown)


max-k:
@patkivikram could you approve this ?
Anvil is one of my teammates and we did this together.
Thank you very much.

patkivikram (approved):

PR #68, opened 2020-12-22
Adding threaded producer

patkivikram:

## Description

that version series.

### DIFF
--- a/faust/agents/agent.py
+++ b/faust/agents/agent.py
@@ -1,5 +1,6 @@
 """Agent implementation."""
 import asyncio
+import sys
 import typing
 from contextlib import suppress
 from contextvars import ContextVar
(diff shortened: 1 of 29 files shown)


max-k:
@patkivikram could you provide a proper description for this ?

faust/agents/agent.py
```diff
@@ -658,7 +658,7 @@ async def _prepare_actor(self, aref: ActorRefT, beacon: NodeT) -> ActorRefT:
         else:
             # agent yields and is an AsyncIterator so we have to consume it.
             coro = self._slurp(aref, aiter(aref))
-        task = asyncio.Task(self._execute_actor(coro, aref), loop=self.loop)
+        task = asyncio.Task(self._execute_actor(coro, aref), loop=self.loop, name=f'{str(aref)}-{self.channel.get_topic_name()}')
```
JeffreyGreenII, line 661: Why are we naming these tasks?

faust/transport/consumer.py
```diff
@@ -830,12 +830,11 @@ async def _commit_livelock_detector(self) -> None:  # pragma: no cover
         await self.sleep(interval)
         async for sleep_time in self.itertimer(interval, name="livelock"):
             if not self.app.rebalancing:
-                await self.verify_all_partitions_active()
+                await self.app.loop.run_in_executor(None, self.verify_all_partitions_active)
```
JeffreyGreenII, line 833: Does this prevent the commit livelock detector from failing when the main thread is blocked during state rebuilds? Otherwise, what's the purpose of making this run_in_executor on a new thread instead of asynchronously processing the `verify_all_partitions_active`?

faust/agents/agent.py
patkivikram: For the asyncio loop to print the stack trace

faust/types/settings/settings.py
```diff
@@ -1326,6 +1327,20 @@ def producer_request_timeout(self) -> float:
         producer batches expire and will no longer be retried.
         """
 
+    @sections.Producer.setting(
+        params.Bool,
+        version_introduced="1.3.1",
```
patkivikram, line 1332: 0.5.0?

faust/agents/agent.py
MicicFilip: hey @patkivikram, I think that this change affected creating agents that are subscribed to multiple topics at the same time. I tested it inside my project and my interpreter is at 3.9. Basically `self.channel.get_topic_name()` raises an exception because there are more then one topics. Can you please revert your changes to the original commit?

PR #90, opened 2021-01-29
fixing race conditions in opening rocksdb

patkivikram:
Fixes issue with rebalance while trying to open rocksdb


patkivikram:
fixes https://github.com/faust-streaming/faust/issues/91

PR #96, opened 2021-02-05
Fix missing sensor and rebalance issues

patkivikram:

## Description

Fix all the race conditions in rebalance and recovery

### DIFF
--- a/faust/stores/rocksdb.py
+++ b/faust/stores/rocksdb.py
@@ -1,5 +1,6 @@
 """RocksDB storage."""
 import asyncio
+import gc
 import math
 import shutil
 import typing
@@ -332,6 +333,7 @@ class Store(base.SerializedStore):
         assigned: Set[TP],
         revoked: Set[TP],
         newly_assigned: Set[TP],
+        generation_id: int = 0,
     ) -> None:
         """Rebalance occurred.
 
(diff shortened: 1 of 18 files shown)


bobh66 (approved):
Looks good to me

faust/stores/rocksdb.py
```diff
@@ -377,26 +385,40 @@ async def assign_partitions(self, table: CollectionT, tps: Set[TP]) -> None:
...
                 # side effect: opens db and adds to self._dbs.
+                self.logger.info(
+                    f"opening partition {partition} for gen id "
+                    f"{generation_id} app id {self.app.consumer_generation_id}"
+                )
                 return self._db_for_partition(partition)
             except rocksdb.errors.RocksIOError as exc:
                 if i == max_retries - 1 or "lock" not in repr(exc):
                     # release all the locks and crash
+                    self.log.warning(
```
jkgenser, line 409: Why is it `self.log` here but `self.logger` above?

jkgenser (approved):
I can confirm that scale up and scale down with 0 standby replicas and one shared data dir works with this setup. 

It also requires `faust-streaming-rocskdb==0.8.0`. Probably will experience rocksdb db locks if you try the shared data dir approach with `python-rocksdb==0.7.0`

patkivikram:
Also implements https://github.com/faust-streaming/faust/issues/98

Issue #93, opened 2021-02-04
Rebalance can leave some workers with no partitions


bobh66:
## Checklist

- [X] I have included information about relevant versions
- [X] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

When the number of partitions for an agent topic is not evenly divisible by the number of workers, the PartitionAssignor can leave one or more workers with no partitions assigned.  This wastes CPU and memory resources, as well as causing other workers to carry a heavier load and potentially reduces throughput.

The initial deployment will distribute the partitions across all workers.  For example, a topic with 100 partitions will be spread across 40 workers with 20 workers having three partitions and 20 workers having 2 partitions.  The CopartitionedAssignor will calculate the (maximum) capacity for each worker to be 3 partitions, using the formula ceil(num_partitions / num_workers).

Now if one worker gets rebooted or leaves the group for any reason, a rebalance is triggered and the 2 or 3 partitions for that worker get moved to other workers that have 2 partitions, leaving 22 or 23 workers with 3 partitions and 17 or 18 workers with 2 partitions (100 partitions across 39 workers).

When the rebooted worker recovers and rejoins the group, it will probably not receive any partitions because there are no "extra" partitions on any of the workers.  The maximum capacity is still 3, no worker has more than 3 partitions, so no partitions are "available" for assignment.  Rebooting the worker has no impact for the same reason.

The worker with no partitions will leave the consumer group after 5 minutes as the aiokafka Fetcher has been idle due to no assignment, which means that future rebalances of the group will NOT include this consumer/worker, and it will be idle forever, or until the group is redeployed.

## Expected behavior

Partitions should be assigned to all available workers, as balanced as possible.

## Actual behavior

Workers can receive no partition assignment and leave the group.

## Full traceback

```pytb
Paste the full traceback (if there is any)
```

# Versions

* Python version - 3.7
* Faust version - 0.4.6
* Operating system - Centos
* Kafka version N/A
* RocksDB version (if applicable) N/A

jkgenser:
@bobh66 : Just wondering, why did you revert your changes to fix the partition assignment problems raised in this issue?

bobh66:
[...]

PR #97, opened 2021-02-09
improve partition assignment balance (#93)

bobh66:

## Description

Modifies the partition assignment logic to produce a more balanced distribution of partitions across clients.  Specifically:

- when table_standby_replicas is set to 0, and the number of partitions is >= the number of clients, on a re-assignment the partitioner will calculate the "minimum" capacity for each client, the smallest number of partitions that each client should have in a balanced distribution, and remove partitions from existing clients that have more than the minimum number.  This is to ensure that there are enough unassigned partitions available to assign to all clients.

- the partitioner will sort the "candidate" clients by the number of (active or standby) partitions that each client has, and then round-robin over that set, so that the clients with the smallest number of partitions get new partitions added to them before clients with larger numbers of partitions.  This only works reliably when table_standby_replicas is 0, because there is logic that will "promote" standby partition assignment to active, which preempts the balancing logic.

### DIFF
--- a/faust/assignor/copartitioned_assignor.py
+++ b/faust/assignor/copartitioned_assignor.py
@@ -1,6 +1,6 @@
 """Copartitioned Assignor."""
 from itertools import cycle
-from math import ceil
+from math import ceil, floor
 from typing import Iterable, Iterator, MutableMapping, Optional, Sequence, Set
 
 from mode.utils.typing import Counter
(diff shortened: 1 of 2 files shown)


jkgenser:
@bobh66: Super excited for this. I came across this recently when I was scaling from 1-8 workers for an 8 partition topic. It's unfortunate that workers 5, 6, and I think 7 do not get a partition. Only once 8/8 workers are up does the load split to 1 partition per worker. This will enable more granular scaling with no standby replicas. Also separately, it looks like in the next release, we'll be able to use one shared rocksdb file for multiple workers which reduces the need for standby replicas in this config. 

Of course, people with standby replicas in a sharded environment might want to use them.

bobh66:
@jkgenser Note that the changes will only remove partitions from existing pods when table_standby_replicas is 0, and the default is 1, so you have to "activate" this change by setting it to 0.  If you are using tables then I would not set table_standby_replicas to 0 unless you don't need the table data to persist across a restart or be accessed by multiple pods.

There is a lot more complexity involved when trying to combine the need for balance with the need for "sticky" and "standby" partitions, so this is the first step to handle the simple case where no standby partitions are involved.

jkgenser:
Makes sense. I'm planning on having multiple workers share the same DATA_DIR and bind mind the same dir to all workers. Even though only one worker will own a given rocksdb-partition at a time. This would be the primary case to use your new assignment logic I think

patkivikram (approved):

PR #102, opened 2021-02-15
fix race (out of order) in flushing topics (wrong state is stored in changelog)

trauter:
## Description

This PR fixes a race condition in writing messages to topics that resulted in a violation of the ordering guarantee (especially changelog topics).

What happened before:
The producer buffer is filled with messages. 
_handle_pending is taking one message from the buffer at a time (Participant 1).
If the buffer grows flush_atmost is called which itself starts taking messages from the buffer and sends them (Participant 2).

In rare cases, both participants popped message (e.g. MSG1, then MSG2), but the messages are written in reverse order (e.g. MSG2, then MSG1) to the topic.
In case of changelog topics, this results in wrong states if both messages have the same key.

The fix:
Basically, flush simply waits until the buffer is empty.

### DIFF
--- a/faust/transport/producer.py
+++ b/faust/transport/producer.py
@@ -56,10 +56,11 @@ class ProducerBuffer(Service, ProducerBufferT):
         """Flush at most ``n`` messages."""
         flushed_messages = 0
         while True:
-            if self.state != 'running' and self.size:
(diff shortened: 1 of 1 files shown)


trauter:
have to fix tests before..

PR #104, opened 2021-02-16
Aerospike implementation in state stores

patkivikram:

## Description

This PR adds support for aerospike as a supported platform for state stores


faust/stores/__init__.py
```diff
@@ -10,6 +10,7 @@
 STORES: FactoryMapping[Type[StoreT]] = FactoryMapping(
     memory="faust.stores.memory:Store",
     rocksdb="faust.stores.rocksdb:Store",
+    aerospike="faust.stores.aerospike:AeroSpikeStore",
```
JeffreyGreenII, line 13: I'm still learning about Aerospike, but does it extend an abstract datastore format that we can use here so that users can _choose_ to use Aerospike as an implementation? (Eg. MariaDB v. MySQL)

However, I don't know if that's realistic when it comes to datastores compared to services.

faust/stores/aerospike.py
```diff
@@ -0,0 +1,198 @@
...
+    import aerospike
+except ImportError:  # pragma: no cover
+    aerospike = None  # noqa
+
+from yarl import URL
+
+from faust.stores import base
+from faust.types import TP, AppT, CollectionT
+
+if typing.TYPE_CHECKING:  # pragma: no cover
```
JeffreyGreenII, line 14: I'm not a Pythonic veteran, so this may be best practice, but why do we have the `typing.TYPE_CHECKING` bool check twice for different imports?

faust/stores/aerospike.py
```diff
@@ -0,0 +1,198 @@
...
+        try:
+            self.client = AeroSpikeStore.get_aerospike_client(options)
+            self.namespace = options.get("namespace", "")
+        except Exception as ex:
+            self.logger.error(f"Error configuring aerospike client {ex}")
+            raise ex
+        super().__init__(url, app, table, **kwargs)
+
+    @staticmethod
+    def get_aerospike_client(aerospike_config: Dict[Any, Any]) -> Client:
```
JeffreyGreenII, line 60: I think I'm seeing a typing discrepancy between the function parameters for `AeroSpikeStore::__init__(options)` and `AeroSpikeStore::get_aerospike_client(get_aerospike_client)`. The `__init__` defaults to `None` for `options`, but `options` is then used for the `aerospike_config` parameter of `get_aerospike_client` which is _not_ optional.

faust/stores/aerospike.py
```diff
@@ -0,0 +1,198 @@
...
+        else:
+            client = aerospike.client(aerospike_config)
+            try:
+                client.connect(None, None)
+                aerospike_client = client
+                return client
+            except Exception as e:
+                raise e
+
+    def _get(self, key: bytes) -> Optional[bytes]:
```
JeffreyGreenII, line 73: Are _all_ keys always going to be `bytes`?
It seems like according to the Aerospike docs, we could make this a `str`, `int`, or `byte array`: https://www.aerospike.com/apidocs/python/client.html#aerospike-key-tuple

May be nice to make this flexible now, but no problem making it bytes now.

faust/stores/aerospike.py
```diff
@@ -0,0 +1,198 @@
...
+                return bins[self.BIN_KEY]
+            return None
+        except aerospike.exception.RecordNotFound as ex:
+            self.log.debug(f"key not found {key} exception {ex}")
+            raise KeyError(f"key not found {key}")
+        except Exception as ex:
+            self.log.error(f"Error in set for table {self.table_name} exception {ex}")
+            raise ex
+
+    def _set(self, key: bytes, value: Optional[bytes]) -> None:
```
JeffreyGreenII, line 87: Same comment here about the `key` being forced to `bytes`

faust/stores/__init__.py
patkivikram: you can plugin your custom adapter using this framework

faust/stores/aerospike.py
patkivikram: no this does not force applications to include aerospike as a dependency even if they dont intend to use it. The tests mock this module and hence we dont need to install aerospike as a library to run tests

faust/stores/aerospike.py
patkivikram: good call

faust/stores/aerospike.py
patkivikram: this is a SerializedStore, so models are byte encoded/decoded in the `Table` framework

faust/stores/aerospike.py
patkivikram: same as above. This layer just deals with bytes

patkivikram:
Fixes https://github.com/faust-streaming/faust/issues/114

PR #106, opened 2021-02-16
Fix iterating over keys when changelog topic is set.

aoberegg:
## Description

Changes the RocksDB store to check the assigned topic name for a changelog, rather than the auto-generated one. If the changelog uses a manually-created topic, using the auto-generated name causes keys(), values(), items() on the Table object that this store backs to return no data, since the changelog names never match.

Resolves https://github.com/robinhood/faust/issues/654
original author: https://github.com/robinhood/faust/pull/655


g-clef:
You'll need to change one of the tests also. See https://github.com/faust-streaming/faust/pull/41/files .

patkivikram (approved):

PR #134, opened 2021-04-18
fix tests directory name in Makefile

taybin:
## Description

The Makefile still refers to the older `t` directory instead of `tests.  Updates that.


patkivikram (approved):

taybin:
@patkivikram please re-approve so it can retest the formatting.

PR #143, opened 2021-04-30
enable the test_aiokafka.py tests

taybin:
## Description

This fixes most of the tests in TestProducer in tests/unit/transport/drivers/test_aiokafka.py.  I found that before use of a Producer inside a transaction, I needed to both call `producer.begin_transaction(<tid>)` and insert the mocked transaction Producer into `producer.transaction_producers[]`.  There is probably a cleaner way of doing this without mucking around inside private attributes, but this seemed like a good place to start a conversation about it.

There were some tests referencing the older `MultiTXNProducer`, which I marked for skipping.  I also skipped the tests around `Producer._new_producer()`, for now.

PR is in draft mode until we're happy with the style of testing and more tests are re-enabled.


taybin:
The tests around `MultiTXNProducer` that I marked as obsolete are now deleted along with some dead code that was commented out.

taybin:
This enables the TestProducer tests, after cleaning them up.  *Something* was putting a Mock inside `Producer._transaction_producers` and I have no idea what it is, but clearing it out after generating a new Producer inside the fixture got the tests working without having to fiddle inside `_transaction_producers` each time.

I'd like to get this merged and then work on the VEP tests in a different PR.

PR #144, opened 2021-05-03
Complete fix rocksdb

xgamer4:
This completes the rocksdb fix for global tables and use_partitioner tables from https://github.com/faust-streaming/faust/pull/130


taybin:
Is there a unit test that can be added for this?

xgamer4:
> Is there a unit test that can be added for this?

Yeah, good call, looks like there's actually a few scenarios that should be added. I'll get those updated.

xgamer4:
Added unit tests on both _get and _contains that cover the changes in this and #130

taybin:
@xgamer4 Thanks for adding the tests! 👍

patkivikram (approved):

Issue #132, opened 2021-04-16
Makefile develop target broken


james-mchugh:
## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Run `make develop`

## Expected behavior

`make develop` runs properly 

## Actual behavior

Running `make develop` fails because default.txt is not in requirements directory.

## Full traceback

```pytb
(faust) ➜  faust git:(master) make develop
pip install -U -r requirements/default.txt
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements/default.txt'
make: *** [reqs-default] Error 1
```

# Versions

* Python version 3.9.4
* Faust version master
* Operating system Darwin 19.6.0
* Kafka version na
* RocksDB version (if applicable)

taybin:
Good catch.  I fixed this, but then I saw that `make develop` still breaks because it runs `pip install -r requirements/dist.txt`, and that references a non-existent `flakes.txt` file.  Should the flake8 linters be moved to that?  Or should the line be deleted from `dist.txt`?  Or is the `Makefile` not really used anymore?  In that case, maybe it should be trimmed down to what is used or even deleted entirely.

taybin:
@james-mchugh Can you try out my PR and see if that works for you?

wbarnha:
Works for me, this should be safe to close once the aforementioned PR is merged.

PR #145, opened 2021-05-03
fix running `make develop` and restore `pre-commit` tool

taybin:
## Description

This is a work in progress for #132, fixing `make develop`.  As I said in [that issue](https://github.com/faust-streaming/faust/issues/132#issuecomment-831544577),
>  I fixed this, but then I saw that make develop still breaks because it runs pip install -r requirements/dist.txt, and that references a non-existent flakes.txt file. Should the flake8 linters be moved to that? Or should the line be deleted from dist.txt? Or is the Makefile not really used anymore? In that case, maybe it should be trimmed down to what is used or even deleted entirely.


taybin:
This fixes another couple inclusions of files that don't exist.  The Makefile still tries to run `pre-commit`, which isn't listed in any of the requirements files either.

I like `pre-commit` myself, but I have a coworker who doesn't, so I'm hesitant to just add it in as another requirement.

taybin:
After seeing that pre-commit was used in the upstream of faust, I added it back in with support for running `flake8`.  It won't be installed unless you run `make develop`, so that seems okay to me.

I'd like to have it run `black` and `isort` as well, but that'll go in a different PR.

wbarnha:
I don't personally use pre-commit or the Makefile (yet) but I support these changes particularly because it cleans up the requirements files. LGTM.

PR #146, opened 2021-05-07
enable testing against python 3.9

taybin:
## Description

Python 3.9 has been out for a while.  We should test against it to make sure it's supported.


taybin:
Not sure how to add the "Required" field to the check.  Glad it passed without changes though.

PR #152, opened 2021-05-17
FIX Record instances deserialize properly when returned by agent.ask

### LINKED ISSUE #151: deserialization of Record objects incomplete when using rpc calls
**Version** Current master

## Steps to reproduce

```python
#adding.py
# based on examples/agent.py
import faust
from faust import Topic

class Add(faust.Record, serializer='json'):
    a: int
    b: int

class Result(faust.Record,serializer='json'):
    r : int
    summands : Add

class Settings:
    FAUST_KAFKA_BROCKER = 'kafka://localhost:9093'
    KAFKA_BROCKERS = ['localhost:9093']

class EcoTopic(Topic):
    def __init__(self, *args, **kwargs) -> None:
        """
        The default replication for Faust topics is broken, this class changes the default
        replication from 0 to 1 and should be used as replacement for Fausts own topics until this is fixed in the
        library
        """
        if kwargs.get("replicas") == 0:
            kwargs["replicas"] = 1
        super().__init__(*args, **kwargs)

app = faust.App(
    'add_app',
    broker=Settings.FAUST_KAFKA_BROCKER,
    Topic=EcoTopic
)

topic = app.topic('adding', value_type=Add)

@app.agent(topic)
async def adding(stream):
    async for value in stream:
        yield Result(value.a + value.b,value)
```
```python
#send_to_agent.py
import faust
from addition.addition_app import Add,Result, adding

class Settings:
    FAUST_KAFKA_BROCKER = 'kafka://localhost:9093'
    KAFKA_BROCKERS = ['localhost:9093']


async def sum_ask(a, b) -> None:
    add_object = Add(a=a, b=b)
    result = (await adding.ask(add_object))
    print(f"result={result}")
    return result

if __name__ == '__main__':
    r = compute_sum_ask(0,0)
    assert isinstance(r, Result)


```
## Expected behavior
Return value should be of class ```Result```

## Actual behavior
Return value is plain json/dict

##Fix
When fulfilling reply promises the return value should be wrapped in a call to ```maybe_model```

##Details
Faust uses some internal store to Register all created Subclasses of ```Record```, when deserializing some special fields in the
deserialized data are used to detect that this should be actually converted to a ```Record```. The corresponding logic is in ```maye_model```.

tarbaig:
Faust uses an internal model store to keep track of classes deriving from ```Record```.
When sent via Kafka those classes are transparently reconstructed from the deserialized data. The corresponding logic resides in ```faust.models.base.maybe_model```  which must be used to wrap the reply value of ```faust.agents.replies.ReplyConsumer._drain_replies```.

fixes #151

### DIFF
--- a/faust/agents/replies.py
+++ b/faust/agents/replies.py
@@ -12,6 +12,8 @@ from .models import ReqRepResponse
 
 __all__ = ["ReplyPromise", "BarrierState", "ReplyConsumer"]
 
+from ..models import maybe_model
+
 
 class ReplyTuple(NamedTuple):
     correlation_id: str
@@ -181,7 +183,7 @@ class ReplyConsumer(Service):
     async def _drain_replies(self, channel: ChannelT) -> None:
         async for reply in channel.stream():
             for promise in self._waiting[reply.correlation_id]:
-                promise.fulfill(reply.correlation_id, reply.value)
+                promise.fulfill(reply.correlation_id, maybe_model(reply.value))
 
     def _reply_topic(self, topic: str) -> TopicT:
         return self.app.topic(
--- a/tests/unit/agents/test_replies.py
+++ b/tests/unit/agents/test_replies.py
@@ -4,9 +4,9 @@ import json
 import pytest
 from mode.utils.mocks import AsyncMock, Mock
 
+from faust import Record
 from faust.agents.models import ReqRepResponse
 from faust.agents.replies import BarrierState, ReplyConsumer, ReplyPromise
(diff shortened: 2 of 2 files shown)


taybin:
This makes sense to me.  The fix is simple and the unit tests were updated.

tarbaig:
Is there anything I need to improve for a merge? It feels silly to keep a fork just to fix my favorite pet bug.

patkivikram (approved):

PR #174, opened 2021-07-20
fix for reassign Table Keys faust-streaming#171

### LINKED ISSUE #171: Table Recovery deletes newly assigned keys
Hello

I have found an error when restoring the table from the Changelog Topic. 
I have a use case where I want to remove a key from the table and reassign it later. For example, a ship arriving in a port in Germany is added to the table. After it has left Germany, it is to be removed from the table. The ship then sails to Australia and back to Germany. When it arrives again, it has to be added again.

When I remove the ship from my table, it is marked as **null** in the Changelog Topic. When restoring, the app checks for **null** values and puts the key in a set of keys to be deleted. Even if a new value is assigned to this key, the key is still in the set of keys to be deleted and the newly assigned value will be removed from table after Recovery. 

## Steps to reproduce
The following table changelog can be used as an example:
```
table_changelog = [
    (1, "Foo"),
    (2, "Bar"),
    (2, None),
    (2, "Baz")
]
```
## Expected behavior
The table should look like this:
   `table = {1: "Foo", 2: "Baz"} `
## Actual behavior
But the table looks like this
    `table = {1: "Foo"}`

## Source Code

The source code to this Bug can be found here: [faust.stores.memory.py](https://github.com/faust-streaming/faust/blob/master/faust/stores/memory.py), lines 18 to 49.

``` python
def apply_changelog_batch(
    self,
    batch: Iterable[EventT],
    to_key: Callable[[Any], Any],
    to_value: Callable[[Any], Any],
) -> None:
    """Apply batch of changelog events to in-memory table."""
    # default store does not do serialization, so we need
    # to convert these raw json serialized keys to proper structures
    # (E.g. regenerate tuples in WindowedKeys etc).
    to_delete: Set[Any] = set()
    delete_key = self.data.pop
    self.data.update(
        self._create_batch_iterator(to_delete.add, to_key, to_value, batch)
    )
    for key in to_delete:
        delete_key(key, None)

def _create_batch_iterator(
    self,
    mark_as_delete: Callable[[Any], None],
    to_key: Callable[[Any], Any],
    to_value: Callable[[Any], Any],
    batch: Iterable[EventT],
) -> Iterable[Tuple[Any, Any]]:
    for event in batch:
        key = to_key(event.key)
        # to delete keys in the table we set the raw value to None
        if event.message.value is None:
            mark_as_delete(key)
            continue
        yield key, to_value(event.value)
```

## Proposal for a solution

In my opinion, keys should not be marked for deletion during iteration. It is better to skip key-value pairs where the value is **null**. That way, if the key is assigned later, it will still be maintained.

# Versions

* Python 3.7.9
* Faust 0.6.9

PJ-Schulz:
## Description

(Fixes #171)
When a key is removed from a table with table.pop(), a changelog entry is sent with a raw null value.

Before this change, on Table recovery, once a key was set to the raw null value, it would be deleted after iterating over the entire Changelog topic.
Now, before deletion, the value for this key is checked to see if it is really null. Otherwise it was assigned to a new value at a later time and will not be deleted

### DIFF
--- a/faust/stores/memory.py
+++ b/faust/stores/memory.py
@@ -31,7 +31,7 @@ class Store(base.Store):
             self._create_batch_iterator(to_delete.add, to_key, to_value, batch)
         )
         for key in to_delete:
-            # If the key was assigned a value again at a later time, it will not be deleted.
+            # If the key was assigned a value again, it will not be deleted.
             if not self.data[key]:
                 delete_key(key, None)
 
--- a/faust/stores/memory.py
+++ b/faust/stores/memory.py
@@ -31,7 +31,9 @@ class Store(base.Store):
             self._create_batch_iterator(to_delete.add, to_key, to_value, batch)
         )
         for key in to_delete:
-            delete_key(key, None)
+            # If the key was assigned a value again at a later time, it will not be deleted.
+            if not self.data[key]:
+                delete_key(key, None)
 
     def _create_batch_iterator(
         self,
@@ -45,7 +47,6 @@ class Store(base.Store):
             # to delete keys in the table we set the raw value to None
             if event.message.value is None:
                 mark_as_delete(key)
-                continue
             yield key, to_value(event.value)
 
     def persisted_offset(self, tp: TP) -> Optional[int]:
--- a/tests/unit/stores/test_memory.py
+++ b/tests/unit/stores/test_memory.py
@@ -33,6 +33,15 @@ class Test_Store:
 
         assert to_key() not in store.data
 
+    def test_apply_changelog_batch__deletes_key_and_reassign_it(self, *, store):
+        self.test_apply_changelog_batch__deletes_key_for_None_value(store=store)
+
+        events = [self.mock_event(value=value) for value in ("v1", None, "v2")]
+        to_key, to_value = self.mock_to_key_value(events[0])
+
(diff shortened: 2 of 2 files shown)


faust/stores/memory.py
```diff
@@ -31,7 +31,9 @@ def apply_changelog_batch(
             self._create_batch_iterator(to_delete.add, to_key, to_value, batch)
         )
         for key in to_delete:
-            delete_key(key, None)
+            # If the key was assigned a value again at a later time, it will not be deleted.
```
patkivikram, line 34: Could you add a quick unit test to test this?

faust/stores/memory.py
```diff
@@ -31,7 +31,9 @@ def apply_changelog_batch(
...
+            if not self.data[key]:
```
patkivikram, line 35: Is this an issue only for memory store? Rocksdb store does not have this issue AFAIK

PJ-Schulz: The issue is only for memory. When I look into the code for Rocksdb I don't recongnize that there can be the same error

faust/stores/memory.py
PJ-Schulz: I have added a test for this case. Until now there was no test for this use case.

patkivikram (approved):

PR #196, opened 2021-10-14
Fix canonical URL when setting either port of host from the CLI

ran-ka:
## Description

A small bug fix to set the right canonical_url when specifying web_port or web_host using the CLI


ran-ka:
@ondrejchmelar can you please see if this PR resolves your issue:
[](https://github.com/faust-streaming/faust/pull/210)

ondrejchmelar:
@ran-ka Yep, that seems right, I would expect it to behave the same way as `web-port`. Thanks!

Issue #202, opened 2021-10-20
Aerospike RuntimeExceptions must be retried


patkivikram:
## Checklist

- [x ] I have included information about relevant versions
- [ x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

In case of runtime exceptions from the aerospike client, the faust app today crashes the app so that the incoming message is not acknowledged

## Expected behavior
This should be a new feature to retry the aerospike operation on a runtime exception

# Versions
0.6.9

* Python version 3.9
* Faust version 0.6.9
* Operating system Centos75
* Kafka version 2.4.1
* RocksDB version (if applicable)

PR #203, opened 2021-10-20
adding new feature to retry on Aerospike runtime exceptions Issue#202

patkivikram:

## Description

Fixes https://github.com/faust-streaming/faust/issues/202


taybin:
Did you consider using a library like `tenacity`?  I've used that in my faust client and it works well.  I know it expands the requirements, but if we want to start having retry code elsewhere, it might be better to use something like that than re-implementing it.

PR #204, opened 2021-10-21
Remove outdated colorclass library

taybin:
## Description

removes `colorclass` requirement and uses of the library.  That library has not been updated since 2016 and no longer works with python-3.10.  There are other options, but they don't easily fit into the faust wrappers around it.  Better to remove it and if necessary, add in a different library later on.


mxcoder (approved):
LGTM

patkivikram (approved):

PR #206, opened 2021-10-23
Feature/support swagger

ran-ka:
## Description

Fix support for Swagger UI

### Goal
Use Swagger and interact with Faust APIs

### Bugs
There were two issues that prevented the existing code from working with Swagger
1) Handler wrapping (_wrap_into_asyncdef)
2) Use of * when adding routes

### Solution
1) Pass __docs__ to the wrapper handler
2) Detect which actual methods are in use, and only create routes for them

### Demo
I added the swagger docs to the existing internal Faust endpoint

image

### Testing
Tested with library:  aiohttp_swagger

```python
from aiohttp_swagger import *
class App(faust.App):
    def on_webserver_init(self, web):
        setup_swagger(
            web.web_app,
            title="Faust test app",
            description = "A simple faust app for internal testing",
            swagger_url="/docs",
            api_version="1.0.3",
            ui_version=2,
        )       

```

### Issues
HEAD endpoint: appears in Swagger, I can remove it, but didn't want to break the existing code
aiohttp-swagger3: Has a different interface, so for now I decided not to support it, but if this is of interest to someone, happy to have a chat

### DIFF
--- a/tests/unit/web/drivers/test_aiohttp.py
+++ b/tests/unit/web/drivers/test_aiohttp.py
@@ -314,7 +314,8 @@ class Test_Web:
         )
 
         web.web_app.router.add_route.assert_has_calls(
-            [call(method, "/foo/", ANY) for method in NON_OPTIONS_METHODS]
+            [call(method, "/foo/", ANY) for method in NON_OPTIONS_METHODS],
+            any_order=True,
         )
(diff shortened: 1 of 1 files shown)


patkivikram:
@ran-ka this should work with any endpoint extending View correct?

Issue #216, opened 2021-11-11
Recovery without tables resumes from kafka before seeking to committed offsets


ekerstens:
## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

- Stateless application with a single agent handling one topic with no tables
- Run multiple nodes of the agent 
- Log the offset handled by the agent and delta between this offset and the previous offset
- Trigger a rebalance by destroying and replacing a node
- Some nodes report an offset delta > 1 and the processor skips some events

## Expected behavior

During a rebalance, an agent which is assigned a topic partition that was already assigned before the rebalance should continue from where it left off. 

## Actual behavior

Instead, in the middle of the rebalance the agent gets a new event which skips offsets. In some cases, the offset then jumps back, but in others the node just keeps going from the skipped offset and never seeks back to where it was before the rebalance. 

# Versions

* Python version: 3.8.8
* Faust version: 0.6.10 
* Operating system: Alpine Linux v3.13
* Kafka version:

patkivikram:
do you have the stack of what your logs are saying?

ekerstens:
Below are some logs produced by our processor which highlight the processor. On each message produced by an agent, we produce a log of the form

```python-traceback
agent=Presence; partition=n; previous_offset=n; current_offset=n; offset_delta=n
```

Here are some logs that showcase the problem
```python-traceback
2021/11/16 21:30:35.805 WARNI module=di_kafka_presence_processor_v3_presence.utils.log_offsets Message=agent=Presence; partition=5; previous_offset=2447291; current_offset=2447292; offset_delta=1
2021/11/16 21:30:35.827 INFO  module=vap_kafka_event.faust.app.base._on_partitions_revoked Message=[^-VapFaustApp]: ON PARTITIONS REVOKED
2021/11/16 21:30:35.835 INFO  module=faust.transport.drivers.aiokafka.consumer.wait_empty Message=[^--Consumer]: STILL WAITING FOR ALL STREAMS TO FINISH
2021/11/16 21:30:35.836 INFO  module=faust.transport.drivers.aiokafka.consumer.wait_empty Message=[^--Consumer]: WAITING FOR 19 EVENTS
2021/11/16 21:30:38.784 WARNI module=aiokafka.consumer.group_coordinator.group_coordinator._do_heartbeat Message=Heartbeat failed for group xxgroupxx because it is rebalancing
... 
[...]

PR #217, opened 2021-11-11
Fix Recovery._resume_streams

ekerstens:
## Description

Fixes https://github.com/faust-streaming/faust/issues/216 

During a rebalance, faust implements multiple branches for recovery to resume processing such as in `Recovery.on_recovery_completed` and `Recovery._restart_recovery`  which roughly follow the pattern of:

  1. `self.log.info("Seek stream partitions to committed offsets.")`
  1. `consumer.perform_seek()`
  1. `self.log.dev("Resume stream partitions")`
  1. `consumer.resume_partitions(`
  1. `consumer.resume_flow()`
  1. `self.app.flow_control.resume()`

However, when the application doesn't use any tables, `_restart_recovery` calls `_resume_streams` which follows the pattern:

  1. `consumer.resume_flow()`
  1. `app.flow_control.resume()`
  1. `self.log.info("Seek stream partitions to committed offsets.")`
  1. `consumer.perform_seek()`
  1. `self.log.dev("Resume stream partitions")`
  1. `consumer.resume_partitions(`

In this case, `app.flow_control.resume()` is called before `consumer.perform_seek()`. This can cause some race conditions where the consumer starts fetching messages from where it left off before queues were cleared by `app.flow_control.suspend()`. Since `suspend` clears queues, if the consumer starts fetching again before seeking back to the last committed offset, a gap appears in the processing. 

This change brings the order of calls in `_resume_streams` in line with elsewhere in the class.

### DIFF
--- a/faust/tables/recovery.py
+++ b/faust/tables/recovery.py
@@ -304,10 +304,6 @@ class Recovery(Service):
         app = self.app
         consumer = app.consumer
         await app.on_rebalance_complete.send()
-        # Resume partitions and start fetching.
-        self.log.info("Resuming flow...")
-        consumer.resume_flow()
-        app.flow_control.resume()
         assignment = consumer.assignment()
         if self.app.consumer_generation_id != generation_id:
             self.log.warning("Recovery rebalancing again")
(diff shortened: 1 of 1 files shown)


patkivikram:
is this ready for review?

ekerstens:
I am working on some validation on an application to make sure the entire issue is solved before marking the PR as ready for review.

ekerstens:
@patkivikram I've found a separate issue where an event can get stuck in the conductor during a rebalance but will open a separate issue/PR for that, this is ready for review.

patkivikram:
sounds good! Are you on the faust slack channel?

patkivikram (approved):

PR #242, opened 2021-12-15
Take with timestamp

nicor88:
# Add take_with_timestamp

## Description

I'm currently using the take method of a stream in production. I have some cases where I would like to have not only the batch of values but a batch of values with the timestamp of when the message was inserted into kafka.

This PR adds a new method called take_with_timestamp, it works exactly like take, but it adds the Kafka timestamp to each returned value.

I tested in a local environment and seems working as expected, the real difference with take is this: 
```
value[timestamp_field_name] = event.message.timestamp
```
the name of the timestamp must be passed as an argument, to force to pass the name, some people might like kafka_timestamp, some event_timestamp, I preferred to leave it generic.

### DIFF
--- a/faust/streams.py
+++ b/faust/streams.py
@@ -392,8 +392,9 @@ class Stream(StreamT[T_co], Service):
             self.enable_acks = stream_enable_acks
             self._processors.remove(add_to_buffer)
 
-    async def take_with_timestamp(self, max_: int, within: Seconds,
-                                  timestamp_field_name: str) -> AsyncIterable[Sequence[T_co]]:
+    async def take_with_timestamp(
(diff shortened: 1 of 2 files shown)


nicor88:
@patkivikram and @ekerstens I saw that you are pretty active, could you have a look? 🙏

taybin:
The code looks clean to me, but I think for a user facing function, there really needs to be a unit test of some kind for it.

nicor88:
@taybin could you have a look again? I added 3 tests, should cover the most essentials parts.

nicor88:
@patkivikram if you have some time to look into that would be amazing, it's pretty trivial. Many thanks.

patkivikram (approved):

PR #253, opened 2022-01-06
Stop fetching when flow stops

ekerstens:
## Description
Changes
1. During recovery, resume the consumer after the app. When the app flow_control resumes it clears all queues, and resuming the consumer after the flow control makes sure that we don't write data to the app which is then discarded.
2. Added logic to the consumer to cancel the fetching coroutine during a rebalance. 
3. Moved the wait for flow to resume from `consumer._wait_next_records` to `consumer.getmany` to avoid a lock during rebalance
4. During a rebalance, the app waits for the consumer to cancel fetching before continuing, to avoid potential issues of discarding a fetch which occurs after assignment.

### DIFF
--- a/faust/transport/consumer.py
+++ b/faust/transport/consumer.py
@@ -479,6 +479,7 @@ class Consumer(Service, ConsumerT):
         self.can_resume_flow = Event()
         self.suspend_flow = Event()
         self.not_waiting_next_records = Event()
+        self.not_waiting_next_records.set()
         self._reset_state()
         super().__init__(loop=loop or self.transport.loop, **kwargs)
         self.transactions = self.transport.create_transaction_manager(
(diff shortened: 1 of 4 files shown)


patkivikram (approved):

faust/transport/consumer.py
```diff
@@ -753,10 +764,19 @@ async def _wait_next_records(
...
+            getmany = self._getmany(
                 active_partitions=active_partitions,
                 timeout=timeout,
             )
+            wait_results = await self.wait_first(getmany, suspend_flow)
+            for coro, result in zip(wait_results.done, wait_results.results):
+                # Ignore records fetched while flow was suspended
+                if coro is suspend_flow:
+                    records = {}
+                    break
```
patkivikram, line 777: Should this be continue?

ekerstens: Reason for break is that even if the fetcher has completed (I have seen cases where both happen around the same time) to ignore the result and return nothing

PR #255, opened 2022-01-08
Make optional types explicit

dbrattli:
## Description

Make optional types explicit. A past version of PEP-484 allowed type checkers to assume an optional type when the default value is None, as in this code:

```py
def handle_employee(e: Employee = None): ...
```
This would have been treated as equivalent to:
```py
def handle_employee(e: Optional[Employee] = None) -> None: ...
```
This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit. For more info see: https://www.python.org/dev/peps/pep-0484/#id29


rhuanbarreto:
@patkivikram can someone approve to run workflows in this PR?

patkivikram (approved):

PR #272, opened 2022-02-22
Producer: Rewrite non enabled unit tests

cbrand:
## Description

During the implementation of https://github.com/faust-streaming/faust/pull/270 I realized that a couple of unit tests inside of the tests/unit/transport/test_producer.py doesn't do anything due to early return statements.

This PR adjusts all the tests to check for expected behavior which did not run correctly before.

Open Question is the `flush_atmost` function. Due to the event set wait method, it cannot be guaranteed that during the flush operation that there are at max the max_messages of sent events being sent. According to the function name and the previous test, however, this is the intended behavior.

I adjusted the test accordingly and just expect that messages are sent out, as there doesn't seem to be a reasonable primitive to fulfill the existing expectation within the test.

### DIFF
--- a/faust/__init__.py
+++ b/faust/__init__.py
@@ -23,7 +23,7 @@ import sys
 import typing
 from typing import Any, Mapping, NamedTuple, Optional, Sequence, Tuple
 
-__version__ = "0.8.3"
+__version__ = "0.8.5"
 __author__ = "Robinhood Markets, Inc."
 __contact__ = "schrohm@gmail.com, vpatki@wayfair.com"
 __homepage__ = "https://github.com/faust-streaming/faust"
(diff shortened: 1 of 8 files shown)


patkivikram (approved):

cbrand:
@patkivikram hopefully fixed the formatting issues (i only ran black on the faust repository and not the tests repository accidentally).

patkivikram:
@cbrand can you resolve the conflicts?

cbrand:
Yes, will do and I will get back to you.

cbrand:
@patkivikram done. Please check. I also realized that it seems that the `ThreadedProducer` isn't tested at all as it would currently fail in Python 3.6 (as it uses asyncio.create_task). See: https://github.com/faust-streaming/faust/pull/277 for a potential solution there.

PR #288, opened 2022-03-24
Improve monitor performance

cbrand:
## Description

Cache via a `WeakKeyDictionary` the str names of tasks and streams. This greatly reduces iterations as this code path takes 50% of the iteration loop on streams in the _py_aiter function in my use case.

*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://faust.readthedocs.io/en/master/contributing.html).

PR #289, opened 2022-03-24
Avoid crash in recovery by pessimistically looking up the earliest key

### LINKED ISSUE #73: RocksDB recovery crashes with KeyError
## Checklist

- [ ] I have included information about relevant versions
- [ ] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

We created a simple test for Faust streaming and ran two or more parallel workers, each using own RocksDB.

Under high traffic, stopping one worker sometimes triggers crashes in the other one.

## Expected behavior

Workers can be stopped, and restarted, and even in the unfortunate even they crash, it should not impact other workers operating in parallel.

## Actual behavior

Other workers may crash.

Tell us what happened instead.

## Full traceback

```
[^---Recovery]: Crashed reason=KeyError(TP(topic='<our-topic>', partition=1)) 
```
```pytb
  File "<our venv>/lib/python3.8/site-packages/mode/services.py", line 802, in _execute_task
    await task
  File "<our venv>/lib/python3.8/site-packages/faust/tables/recovery.py", line 375, in _restart_recovery
    await self._wait(
  File "<our venv>/lib/python3.8/site-packages/faust/tables/recovery.py", line 562, in _wait
    wait_result = await self.wait_first(coro, signal, timeout=timeout)
  File "<our venv>/lib/python3.8/site-packages/mode/services.py", line 715, in wait_first
    f.result()  # propagate exceptions
  File "<our venv>/lib/python3.8/site-packages/faust/tables/recovery.py", line 666, in _build_offsets
    new_value = earliest[tp]
KeyError: TP(topic='<our topic>', partition=1)
```

# Versions

* CPython 3.8.5
* Faust faust-streaming @ git+https://github.com/faust-streaming/faust@fcb6b185105239329fd21e4af0f23e5244a2c8eb
* Ubuntu 20.04.1 LTS
* Kafka 2.6.0
* python-rocksdb 0.7.0

cbrand:
## Description

The earliest_offset function in the consumer seems to not always return all tps which have been passed in but presumabely only the ones it has data for.

After longer runtime it seems that the response MAY return a dict not filled with all topic partitions which have been added. To bypass this and not lead faust to crash, use the .get function on the returned mapping and default to None on return.

I think this error occurs if the request happens during a rebalance looking at the code base of aiokafka, but am not totally sure. This change should however not crash the overall recovery process if this occurs. If a rebalance happens on the next call the recovery will restart but will not hard crash as it does right now.

There also seems to be an old issue which I address in this PR:
(Fixes #73)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,5 @@
 *Note*: Before submitting this pull request, please review our [contributing
-guidelines](https://faust.readthedocs.io/en/master/contributing.html).
+guidelines](https://fauststream.com/en/latest/contributing.html).
 
 ## Description
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -19,7 +19,7 @@ jobs:
       # for example if a test fails only when Cython is enabled
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         use-cython: ["true", "false"]
     env:
       USE_CYTHON: ${{ matrix.use-cython }}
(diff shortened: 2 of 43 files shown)


cbrand:
Cherry picked the flake8 fixes from https://github.com/faust-streaming/faust/pull/288 to allow CI to pass.

PR #304, opened 2022-05-14
Add method to RocksDB for backing up partitions

wbarnha:
I have a specific use-case where I need to actively backup a RocksDB store. I've had success using `rocksdb.BackupEngine` with `faust` but it'd be much more usable if the code was integrated into `faust`.

I should note I've only had success with this using `librocksdb-dev==6.11.4-3` on Ubuntu 22.04 and the python-rocksdb bindings from https://github.com/NightTsarina/python-rocksdb.

If you have interest in this, let me know because I really need this feature.

### DIFF
--- a/faust/stores/aerospike.py
+++ b/faust/stores/aerospike.py
@@ -263,7 +263,9 @@ class AeroSpikeStore(base.SerializedStore):
                 )  # crash the app to prevent the offset from progressing
             raise ex
 
-    async def backup_partition(self, tp: Union[TP, int], flush: bool = True, purge: bool = False, keep: int = 1) -> None:
+    async def backup_partition(
+        self, tp: Union[TP, int], flush: bool = True, purge: bool = False, keep: int = 1
+    ) -> None:
         """Backup partition from this store.
 
         Not yet implemented for Aerospike.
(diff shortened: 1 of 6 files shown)


wbarnha:
I'm satisfied with the changes I have now. If you have any other changes you'd like to see, let me know.

wbarnha:
I forgot to test some last-minute changes earlier. I've tested everything and runs properly now.

wbarnha:
Removed import of `BackupEngine` to pass lint check, should be good now...

wbarnha:
I'm not sure where this should be added, but according to https://fauststream.slack.com/archives/C7FL9CB5E/p1594816114113200?thread_ts=1594815786.109000&cid=C7FL9CB5E, recoveries are automatically triggered when a table is accessed. After testing further, it seems that calling `restore_backup` is equally safe as calling it during initialization or in asynchronous tasks that startup upon app initialization. Still haven't tested periodically restoring backups while the app is initialized since I haven't needed to do that.

wbarnha:
On further testing, it appears that the current release of https://github.com/faust-streaming/python-rocksdb is fine for accessing the backup functions in `librocksdb-dev`. For older versions that are released such as 5.8.8, I got errors along the lines of `rocksdb.error.NotImplementedError`. We may want to add a disclaimer that the backup options are only available in specific versions of librocksdb.

patkivikram (approved):

Issue #312, opened 2022-05-31
Faust commits the wrong offset in case of a gap in acks


ekerstens:
## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

- Create an agent consuming from a topic defined like
```python
counter = 0
async def test_event_loop2(stream: Stream) -> AsyncIterable[None]:
    async for event in stream.noack().events():
        counter += 1
        print(event.message.refcount)
        print(event.message.offset)
        print(app.consumer._acked)
        if not event.message.offset % 5 == 0:
            event.ack()
        elif event.message.offset % 10 == 0: 
            breakpoint()
        
        if counter == 30:
            return
```
- Acked messages should appear in app.consumer._acked, but with gaps for the uncommitted offsets every 5th offset
- The consumer will continue to commit offsets, despite the skipped messages. 

## Expected behavior

Faust should be able to identify a gap between the committed offset and the first acked message to avoid skipping messages. Per the documentation in [consumer._new_offset](https://github.com/faust-streaming/faust/blob/7cc93e8f1e8fc65ce7664c3c229b59c259b05e95/faust/transport/consumer.py#L1074)

>         # We iterate over it until we find a gap
>         # then return the offset before that.
>         # For example if acked[tp] is:
>         #   1 2 3 4 5 6 7 8 9
>         # the return value will be: 10
>         # If acked[tp] is:
>         #  34 35 36 40 41 42 43 44
>         #          ^--- gap
>         # the return value will be: 37

Since the above scenario, we never ack every 5th message, the offset should never continue as there is a gap.

## Actual behavior

- Suppose `app.consumer._acked == [1,2,3,4,6,7]`
- When `Consumer._new_offset` is called for the first time, a gap is identified at 5 and so offset 5 is committed. 
- After this, `app.consumer._acked == [6,7]`
- Even if no more messages are consumed or acked, when `Consumer._new_offset` is called next, there is no gap, so the consumer commits offset 8. 
- However, this is incorrect because in reality, there is still a gap when you consider the current committed offset. 

# Versions

* Python version: Python 3.8.12
* Faust version: faust-streaming==0.8.4

## Traceback/Logs
```bash
──────────┴────────┘
2022/05/31 17:15:19.639 WARNI module=mode.redirect.logging.write Message=1
2022/05/31 17:15:19.640 WARNI module=mode.redirect.logging.write Message=2781
[...]

PR #313, opened 2022-06-01
Faust commits the wrong offset in case of a gap in acks #312

s3563602123:
fix new_offset function edge case #312 

## Description

Add changes to handle gap in the head of acked queue:
- We iterate over it until we handle gap in the head of acked queue, then return the previous committed offset.
            # For example if acked[tp] is:
            #    34 35 36 37
            #  ^-- gap
            # self._committed_offset[tp] is 31
            # the return value will be: 31

### DIFF
--- a/faust/transport/consumer.py
+++ b/faust/transport/consumer.py
@@ -1104,7 +1104,7 @@ class Consumer(Service, ConsumerT):
             #    34 35 36 37
             #  ^-- gap
             # self._committed_offset[tp] is 31
-            # the return value will be: 31
+            # the return value will be None (the same as 31)
             if self._committed_offset[tp]:
                 if min(acked) - self._committed_offset[tp] > 0:
                     return None
--- a/faust/transport/consumer.py
+++ b/faust/transport/consumer.py
@@ -1107,7 +1107,7 @@ class Consumer(Service, ConsumerT):
             # the return value will be: 31
             if self._committed_offset[tp]:
                 if min(acked) - self._committed_offset[tp] > 0:
-                    return self._committed_offset[tp]
+                    return None
 
             # Note: acked is always kept sorted.
             # find first list of consecutive numbers
--- a/faust/transport/consumer.py
+++ b/faust/transport/consumer.py
@@ -1097,6 +1097,18 @@ class Consumer(Service, ConsumerT):
                 acked.extend(gaps)
                 gap_for_tp[:gap_index] = []
             acked.sort()
+
+            # We iterate over it until we handle gap in the head of acked queue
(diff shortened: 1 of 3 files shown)


faust/transport/consumer.py
```diff
@@ -1097,6 +1097,18 @@ def _new_offset(self, tp: TP) -> Optional[int]:
...
+            # We iterate over it until we handle gap in the head of acked queue
+            # then return the previous committed offset.
+            # For example if acked[tp] is:
+            #    34 35 36 37
+            #  ^-- gap
+            # self._committed_offset[tp] is 31
+            # the return value will be: 31
+            if self._committed_offset[tp]:
+                if min(acked) - self._committed_offset[tp] > 0:
+                    return self._committed_offset[tp]
```
ekerstens, line 1110: We should return None here. Returning the currently committed offset would cause us to recommit the same value, but in that case there isn't really a need to commit anything.

e-kosten (approved):

faust/transport/consumer.py
```diff
@@ -1097,6 +1097,18 @@ def _new_offset(self, tp: TP) -> Optional[int]:
...
+
+            # We iterate over it until we handle gap in the head of acked queue
+            # then return the previous committed offset.
+            # For example if acked[tp] is:
+            #    34 35 36 37
+            #  ^-- gap
+            # self._committed_offset[tp] is 31
+            # the return value will be: 31
+            if self._committed_offset[tp]:
+                if min(acked) - self._committed_offset[tp] > 0:
```
patkivikram, line 1109: What if there is gap for that offset in the kafka topic? Do we have to check for that?

ekerstens: I think that should be handled already since the gaps are combined into acked above. @s3563602123 can you add a test case to confirm?

s3563602123: Added test cases to verify this situation

faust/transport/consumer.py
```diff
@@ -1097,6 +1097,18 @@ def _new_offset(self, tp: TP) -> Optional[int]:
...
                 gap_for_tp[:gap_index] = []
             acked.sort()
+
+            # We iterate over it until we handle gap in the head of acked queue
+            # then return the previous committed offset.
+            # For example if acked[tp] is:
+            #    34 35 36 37
+            #  ^-- gap
+            # self._committed_offset[tp] is 31
+            # the return value will be: 31
```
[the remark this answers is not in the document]
s3563602123: Done

patkivikram (approved):

PR #331, opened 2022-07-26
Allow rocksdb to accept read_only option in order to share tables among workers

### LINKED ISSUE #330: Multi-process reads for RocksDB among multiple Faust apps
I'd like to be able to setup multiple workers/Faust Apps that have the same name to read the same Faust Table entries. I may need to make some changes to `python-rocksdb` in order to achieve this.

I understand some of this functionality was added in https://github.com/faust-streaming/faust/issues/98, the note from Ask Solem in https://github.com/robinhood/faust/issues/299#issuecomment-468088562 is no longer true.

Taken from https://github.com/aayushKumarJarvis/rocks-wiki/blob/master/RocksDB-FAQ.md:
```
Q: Can I write to RocksDB using multiple processes?

A: No. However, it can be opened in read-only mode from multiple processes.

Q: Does RocksDB support multi-process read access?

A: RocksDB support multi-process read only process without writing the database. This can be done by opening the database with DB::OpenForReadOnly() call
```

Could we have a Faust app configured to only read from a RocksDB file so that multiple processes could share one RocksDB file and only one process writes to the RocksDB file?

wbarnha:
Closes #330, I've had success by running multiple Faust apps with the same data directory for RocksDB
```
app = faust.App("reuse-this-app-name")
app.GlobalTable('generic-table-name', options={'read_only': True})
```

I'd like to write some test cases/examples before merging this in order to understand how this works better. If this functionality is not intended, feel free to close this immediately.

### DIFF
--- a/faust/stores/rocksdb.py
+++ b/faust/stores/rocksdb.py
@@ -138,7 +138,10 @@ class RocksDBOptions:
 
 
 class Store(base.SerializedStore):
-    """RocksDB table storage."""
+    """RocksDB table storage.
+    Pass 'options={'read_only': True}' as an option into a Table class
+    to allow a RocksDB store be used by multiple apps.
+    """
 
     offset_key = b"__faust\0offset__"
 
--- a/faust/stores/rocksdb.py
+++ b/faust/stores/rocksdb.py
@@ -178,10 +178,7 @@ class Store(base.SerializedStore):
         if not self.url.path:
             self.url /= self.table_name
         self.options = options or {}
-        self.read_only = read_only
-        if "read_only" in self.options:
-            self.read_only = self.options.get("read_only")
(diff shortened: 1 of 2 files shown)


wbarnha:
I've been using https://github.com/NightTsarina/python-rocksdb with Faust. I made a PR for that update in https://github.com/faust-streaming/python-rocksdb/pull/1 so that we can synchronize progress on all forks. Feel free to merge that whenever you feel comfortable.

I just tested this PR with the current release of https://github.com/faust-streaming/python-rocksdb and I can confirm everything works properly. Before using the changes in this PR, I got errors such as `IO error: lock [DB path]/LOCK: Resource temporarily unavailable` from RocksDB. 

It'd be nice if I could make the `read_only` option more accessible.

PR #334, opened 2022-08-05
Correct pypi link, tar filename and path

afasseco:
## Description

Thanks for creating the faust-streaming fork.

We have been using faust for a while and are in process of switching over to faust-streaming. During this I wanted to find latest pypi package and saw the link is still pointing to the faust pypi package, which could be confusing.

So I corrected the link and while doing that also the small untarring process description below that.

Thanks for the good work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,13 +334,13 @@ The following configuration options should be passed in as keys to the options p
 
 ## Downloading and installing from source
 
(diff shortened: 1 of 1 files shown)


wbarnha:
Thanks for the correction! Will merge.

Issue #333, opened 2022-08-02
Suffered data loose when using a rocksdb store when we have an intense write


YiuRULE:
## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Small script to reproduce it.

```
 import faust
 from faust import cli
 
 app = faust.App("example", broker="kafka://localhost:9092", store="rocksdb://")
 
 
 class Record(faust.Record):
     id: int
 
 
 topic = app.topic(
     "count-record",
     value_type=Record,
     partitions=1,
 )
 table = app.Table(
     "count-state",
     default=lambda: 0,
     partitions=1,
 )
 
 
 @app.agent(topic)
 async def myagent(stream):
     record_id = 0
     async for event in stream:
         record_id += 1
         previous_value = table[event.id]
         current_value = previous_value + 1
         table[event.id] = current_value

         print("==============================")
         print(f"Record ID: {record_id}")
         print(f"Previous value: {previous_value}")
         print(f"Current value: {current_value}")
 
 
 
 @app.command()
 async def populate(self: cli.AppCommand) -> None:
     for _ in range(1000000):
         await topic.send(value=Record(id=1), key=str(1))
```

Should be run using `faust --loop=uvloop --app=FOLDER worker` and `faust --loop=uvloop --app=FOLDER populate`.

## Expected behavior

The purpose of this script is simply to count how many event was found for a specific ID, if we run both command only one time, both the line `Record ID` and `Current value` should be identical, as the ID is fix for the script who populate data into kafka.

## Actual behavior

After running the script long enough, our tables came back from a previous state and we suspect if may be related to the changelog. Here a potential trace while running the script locally.

```
[2022-08-02 08:46:57,932] [87369] [WARNING] Record ID: 29079
[2022-08-02 08:46:57,932] [87369] [WARNING] Previous value: 29078
[2022-08-02 08:46:57,932] [87369] [WARNING] Current value: 29079
[2022-08-02 08:46:57,934] [87369] [WARNING] ==============================
[2022-08-02 08:46:57,934] [87369] [WARNING] Record ID: 29080
[2022-08-02 08:46:57,934] [87369] [WARNING] Previous value: 1000
[2022-08-02 08:46:57,934] [87369] [WARNING] Current value: 1001
```

Ideally on the second run, previous value should be equal of 29079 as it was the value set on its previous run, but it come back with a value of 1000.

[...]

PR #335, opened 2022-08-08
Fix recovery of partitions for large volume writes

wbarnha:
Intended to address #333.


wbarnha:
Merging this ASAP for patching a critical issue

PR #337, opened 2022-08-10
missing `f` prefix on f-string

russellpierce:
## Description

Clones https://github.com/robinhood/faust/pull/755/files for faust-streaming to fix https://github.com/robinhood/faust/issues/754#issue-1213465684 - which I didn't create for this repo to avoid clutter for such a small fix.


wbarnha:
Good catch, thanks for the PR!

PR #344, opened 2022-08-29
add noqa: B024 to base classes to pass lint check

wbarnha:
Current lints are failing due to an update in `flake8-bugbear`, this will let our tests pass again.


wbarnha:
It's still failing for Python 3.6+3.7:
```
faust/types/models.py:46:5: B024 _InitSubclassCheck is an abstract base class, but it has no abstract methods. Remember to use @abstractmethod, @abstractclassmethod and/or @abstractproperty decorators.
```

Issue #318, opened 2022-06-16
Offset not committed after partition returned back to worker after 2 rebalances


ostetsenko:
## Checklist

- [ ] I have included information about relevant versions
- [ ] I have verified that the issue persists when using the `master` branch of Faust.

Do we need to sync this fix:
https://github.com/robinhood/faust/issues/681

Looks like I have the same issue with faust==0.8.5

1. After starting I see failed offset
`Fetch offset 1446280 is out of range for partition TopicPartition(topic='raw.tag', partition=2), resetting offset`
2. Stream reset offset
3. Messages were processed successfully
4. After pod restart, I see the same every time.

Any messages like "Sending offset-commit request with.."

wbarnha:
Thanks for pointing to the original issue and its MR in https://github.com/robinhood/faust/pull/701.

Based on what I see in https://github.com/faust-streaming/faust/blob/master/faust/transport/consumer.py#L1124:
```python
            batch = next(consecutive_numbers(acked))
            # remove them from the list to clean up.
            acked[: len(batch)] = []
            self._acked_index[tp].difference_update(batch)
            # return the highest commit offset
            return batch[-1] + 1
```
I think it's possible that we never patched this. There's also another note https://github.com/faust-streaming/faust/commit/136b781bf2825802157427326f45b3b8179e2d0e#r44636889.
`return batch[-1] + 1` was originally `return batch[-1]`, which was changed as a product of the PR in https://github.com/faust-streaming/faust/pull/45.

I know our library is tangentally different from the original project now, but it could be worth trying to integrate and test changes from https://github.com/robinhood/faust/pull/701.

wbarnha:
I tested #345 and I don't see the issue described https://github.com/robinhood/faust/issues/681, but I also tested with v0.8.5 and I did not see the issue persist as well. I see the offsets committing fine. @ostetsenko is your configuration identical to https://github.com/robinhood/faust/issues/681? Could you provide us logs?

wbarnha:
Closing due to inactivity and also because I'm pretty sure this issue has been patched.

patkivikram:
does that offset exist on your topic? Could be that those offsets were cleaned by the broker?

PR #345, opened 2022-08-30
Integrate change from https://github.com/robinhood/faust/pull/701

wbarnha:
Intended to address https://github.com/faust-streaming/faust/issues/318.


wbarnha:
I'm hesitant to merge this because after testing, we may not need this MR.

faust/transport/consumer.py
```diff
@@ -1122,6 +1122,9 @@ def _new_offset(self, tp: TP) -> Optional[int]:
             batch = next(consecutive_numbers(acked))
             # remove them from the list to clean up.
             acked[: len(batch)] = []
+            latest_commit_id = self._committed_offset.get(tp, -1)
```
patkivikram, line 1125: is there a unit test to validate this change?

wbarnha: I don't think so. I think we should write one to make sure this doesn't break functionality.

faust/transport/consumer.py
```diff
@@ -1122,6 +1122,9 @@ def _new_offset(self, tp: TP) -> Optional[int]:
...
+            if acked and latest_commit_id and acked[0] <= latest_commit_id:
+                acked[:1] = []
```
patkivikram, line 1127: isn't this same as acked[0] = []

wbarnha: Yeah, `acked[:1] = []` is the same as `acked[0] = []`, not sure why that syntax was chosen.

wbarnha:
Unless anybody else can prove this MR is needed, I don't think we need to merge this. We already clear the acked buffer after rebalance. Closing.

PR #362, opened 2022-09-16
Update and revamp CHANGELOG.md

wbarnha:
Until we have a way to automatically regenerate CHANGELOG.md files, this is probably a good way to go for now.


wbarnha:
`git-changelog` still doesn't properly make logs for _every_ change, for example I had to manually write the changes in `/requirements` into the v0.8.10 release.

Issue #166, opened 2021-06-29
New error messages seen in faust app with aiokafka 0.7.1


patkivikram:
## Checklist

- [x ] I have included information about relevant versions
- [x ] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Bring up the faust app and let it run for a while. After running for some time the application starts logging the following error messages

```
2021-06-28 10:57:33 [1] [ERROR] faust.transport.drivers.aiokafka [^----AIOKafkaConsumerThread]: Has not committed TP(topic='test_topic', partition=27) at all since worker start (started 5.52 minutes ago).

There are multiple possible explanations for this:

1) The processing of a single event in the stream
   is taking too long.

    The timeout for this is defined by the broker_commit_livelock_soft_timeout setting,
    currently set to 300.0.  If you expect the time
    required to process an event, to be greater than this then please
    increase the timeout.

## Expected behavior

These errors should only happen if the commit offset is not progressing

## Actual behavior
The app is running fine and consumer offsets are actually updated

# Versions

* Python version 3.9
* Faust version 0.6.7
* Operating system Centos75
* Kafka version 2.4.1

**Comment by bluefatter** [comment]
Thanks for fix this. [Party]

**Comment by bitdivision** [comment]
We're still seeing this issue with v0.6.9

**Comment by patkivikram** [comment]
@bitdivision what are you seeing? Is this on a single worker or on multiple?

**Comment by bitdivision** [comment]
We've now ignored all of the following logs in sentry configuration, however we are still seeing the logs for them in 0.6.9:
```
        "[^---AIOKafkaConsumerThread]: Has not committed",
        "[^---AIOKafkaConsumerThread]: Aiokafka has not sent fetch request for",
        "[^---AIOKafkaConsumerThread]: Stream has not started processing",
        "[^---AIOKafkaConsumerThread]: Stream stopped processing, or is slow for",
```

We're running with 6 workers. I can give more details on specific logs if that's helpful?

patkivikram:
can you share the logs you are seeing in 0.6.9 with more than 1 worker? It should not happen with 1 worker I hope :)

popadi:
Any update on this? It still happens and it's hard to tell if it's a real issue or not. I'm using 0.6.10.

richardhundt:
If you run a worker with 1 topic N partitions, it only consumes from a single partition. After 300 seconds the monitor starts complaining that N-1 partitions are idle.

[...]

PR #377, opened 2022-10-06
Fix Aiokafkaconsumer thread tests

wbarnha:
Hopefully this will help us investigate #166.


wbarnha:
Lots of woes here related to opentracing, which is now deprecated: https://github.com/opentracing/specification/issues/163

wbarnha:
After further investigation, there's a lot that needs to be done for fixing `opentracing` usage. I think the move for now is to merge this as-is and then migrate `opentracing` usage over to `opentelemetry` as documented in https://opentelemetry-python.readthedocs.io/en/stable/shim/opentracing_shim/opentracing_shim.html.

patkivikram (approved):

PR #388, opened 2022-10-14
add variable bound agent example

joonjeong:
## Description

Add variable-bound agent example with dead letter handling use case

### DIFF
--- a/examples/bound_agent.py
+++ b/examples/bound_agent.py
@@ -1,0 +1,79 @@
+import logging
+from random import randint
+import time
+from functools import partial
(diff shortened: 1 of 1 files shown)


wbarnha (approved):

wbarnha:
I'd like to test this out for myself before merging it, I have no issues merging this otherwise.

joonjeong:
just run script is enough..!

```bash
faust $ python examples/bound_agent.py worker -l info
```

and you can see like this.

스크린샷 2022-10-21 오전 3 35 41

PR #392, opened 2022-10-23
Add take_events() function

nguyentritai2906:
## Description

This PR is from [robinhood/faust](https://github.com/robinhood/faust/pull/661) adds a new function stream.take_events() which will return events()


wbarnha:
Nice catch, this should hopefully close a lot of ongoing bugs.

wbarnha:
Looks like the code needs to be linted.

wbarnha (approved):
Doesn't break any features and adds functionality, so I consider this a definite approve. LGTM.

PR #407, opened 2022-11-04
Use setuptools_scm to automate versioning

### LINKED ISSUE #406: Use setuptools_scm to automatically detect version
I'm getting tired of manually setting versions. Let's use some library that does it for us.

wbarnha:
Closes #406.

Still not done yet because I need the version in https://github.com/faust-streaming/faust/blob/master/docs/includes/tags.txt to automatically update.

I'm also removing the `publish` script since we're standardizing releases through Github Releases from hereon.

Before merging this, I'd like to use `setuptools_scm` in a non-deprecated way as well, by having it outside of `setup.py` and in `setup.cfg`.


wbarnha:
Apparently usage of `setuptools_scm` in `setup.py` _and_ `setup.cfg` is deprecated, we'd have to make a `pyproject.toml` file.

wbarnha:
Should probably remove `bumpversion` as well.

wbarnha:
https://wbarnha.github.io/faust/introduction.html looks good.

wbarnha:
Before merging this, I should verify releases to PyPi work properly.

wbarnha:
PyPi test release successful. LGTM.

PR #408, opened 2022-11-04
use importlib instead

wbarnha:
We need to remove setuptools_scm from init because the documentation was misleading.


wbarnha:
Apparently this breaks pages builds as indicated in https://github.com/wbarnha/faust/actions/runs/3396191130/jobs/5646978695.

patkivikram (approved):

Issue #419, opened 2022-12-02
Errors with python3.11


krzysieksulejczak:
## Checklist

- [X] I have included information about relevant versions
- [X] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

I tried to switch to python3.11, I used faust-streaming==0.10.0 and mode-streaming==0.3.2. The application started but the agent started throwing errors and eventually died.

## Expected behavior

The agent should process messages without errors.

## Actual behavior

see tracebacks below

## Full traceback

```pytb
[2022-12-02 14:13:47,323] [1] [ERROR] [^--Consumer]: Drain messages raised: TypeError('Passing coroutines is forbidden, use tasks explicitly.')
Traceback (most recent call last):
  File "/app/venv/lib/python3.11/site-packages/faust/transport/consumer.py", line 1200, in _drain_messages
    await self.wait_first(
  File "/app/venv/lib/python3.11/site-packages/mode/services.py", line 769, in wait_first
    f.result()  # propagate exceptions
    ^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/faust/transport/conductor.py", line 274, in on_message
    return await get_callback_for_tp(message.tp)(message)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "faust/transport/_cython/conductor.pyx", line 78, in __call__
  File "/usr/local/lib/python3.11/asyncio/tasks.py", line 415, in wait
    raise TypeError("Passing coroutines is forbidden, use tasks explicitly.")
TypeError: Passing coroutines is forbidden, use tasks explicitly.
[2022-12-02 14:13:47,326] [1] [ERROR] [^---Fetcher]: Crashed reason=TypeError('Passing coroutines is forbidden, use tasks explicitly.')
Traceback (most recent call last):
  File "/app/venv/lib/python3.11/site-packages/mode/services.py", line 853, in _execute_task
    await task
  File "/app/venv/lib/python3.11/site-packages/faust/transport/consumer.py", line 177, in _fetcher
    await consumer._drain_messages(self)
  File "/app/venv/lib/python3.11/site-packages/faust/transport/consumer.py", line 1200, in _drain_messages
    await self.wait_first(
  File "/app/venv/lib/python3.11/site-packages/mode/services.py", line 769, in wait_first
    f.result()  # propagate exceptions
    ^^^^^^^^^^
[...]

PR #420, opened 2022-12-06
fix for issue 419

patkivikram:
## Description
Possible fix for https://github.com/faust-streaming/faust/issues/419

that version series.


wbarnha (changes requested):
I don't this is the place we need to make changes, we should try looking at this?
https://github.com/faust-streaming/faust/blob/ebc7777333885dccf846d99e0b72f07511b02092/faust/transport/conductor.py#L169-L175

wbarnha (approved):

PR #422, opened 2022-12-14
Assign blessed_key to metadata if blessed_key is defined

### LINKED ISSUE #421: Faust doesn't serialize properly the model with a blessed_key
Faust doesn't serialize properly the model with a blessed_key.

## Checklist

- [X] I have included information about relevant versions
- [X] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce
```
class Foo(faust.Record, serializer='json'):
    _blessed_key = "__meta__"
    id: str = "foo"
...

foo = Foo.loads(b'{"id":"foo","__meta__":{"ns":"app.models.Foo"}}')  # Deserialize model object
print(foo.dumps())  # Serialize object
```

## Expected behavior

`
print(foo.dumps())  # b'{"id":"foo","__meta__":{"ns":"app.models.Foo"}}'
`

## Actual behavior

`
print(foo.dumps())  # b'{"id":"foo","__faust":{"ns":"app.models.Foo"}}'
`

## Versions
* Python version: 3.8.10
* Faust version: 0.10.0
* Operating system: Ubuntu 20.04
* Kafka version: 7.3.0 (Confluent docker)

wbarnha:
Fixes #421.

### DIFF
--- a/faust/models/record.py
+++ b/faust/models/record.py
@@ -592,9 +592,7 @@ class Record(Model, abstract=True):  # type: ignore
         payload = self.asdict()
         options = self._options
         if options.include_metadata:
-            payload[self._blessed_key if self._blessed_key else "__faust"] = {
-                "ns": options.namespace
-            }
+            payload[self._blessed_key] = {"ns": options.namespace}
         return payload
 
     def asdict(self) -> Dict[str, Any]:  # pragma: no cover
--- a/faust/models/record.py
+++ b/faust/models/record.py
@@ -592,7 +592,9 @@ class Record(Model, abstract=True):  # type: ignore
         payload = self.asdict()
         options = self._options
         if options.include_metadata:
-            payload["__faust"] = {"ns": options.namespace}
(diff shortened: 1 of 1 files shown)


faust/models/record.py
```diff
@@ -592,7 +592,7 @@ def to_representation(self) -> Mapping[str, Any]:
         payload = self.asdict()
         options = self._options
         if options.include_metadata:
-            payload["__faust"] = {"ns": options.namespace}
+            payload[self._blessed_key] = {"ns": options.namespace}
```
wbarnha, line 595: Apparently `options.allow_blessed_key` is defined in `faust/types/models.py`, do we want to add this as well to check before adding the blessed key?

wbarnha: https://github.com/faust-streaming/faust/blob/ebc7777333885dccf846d99e0b72f07511b02092/faust/models/base.py#L153-L160

Guess we don't need the option here.

Issue #283, opened 2022-03-09
GlobalTable not working when using multiple instances


Hamdiovish:
## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Running the following app in 2 instances, then by pushing a message to one partition, the instance listening to the related partition detect the message and update the GlobalTable, whilst the other instance will not update its version of the GlobalTable and keep showing nothing.

```
import faust
from faust.cli import option

app = faust.App(
    'hello-world',
    broker='kafka://localhost:29092',
    key_serializer='raw',
    value_serializer='raw',
	store="rocksdb://", 
	topic_disable_leader=True,
    )

greetings_topic = app.topic('greetings-topic',partitions=2,key_type=str, value_type=str,internal=True)
greetings_table = app.GlobalTable('greetings-table',partitions=2,key_type=str, value_type=str,default=str)

@app.agent(greetings_topic)
async def greet(greetings):
	"""run: faust -A app worker"""
	async for event in greetings.events():
		k = event.key
		v = event.value
		print(f"processing: {k}: {v} on partition: {event.message}")
		greetings_table[k]=v

@app.command(
	option('--k', type=str, default='a',help='The key.'),
	option('--v', type=str, default='0',help='The value.'))
async def sim_greeting(self, k, v, p):
	"""simulate: faust -A app01 sim-greeting --k a --v 0"""
	await greetings_topic.send(key=k,value=v)

@app.timer(interval=1.0)
async def plot(app):
	for k in greetings_table:
		print(f'[{k}:{greetings_table[k]}]')
```

## Expected behavior

All instances should detect GlobalTable updates.

## Actual behavior

The GlobalTable is not synced across instances.

# Versions

* Python version: 3.7.12
* Faust version: 0.8.4
* Operating system: Ubuntu 18
* Kafka version: 2.8-IV1
* RocksDB version: 0.8.0

Roman1us:
Please check partitioning (what instance consume partition) and rocksdb storage use different paths on instances

Hamdiovish:
@Roman1us regarding paths, i’m using different —datadir and —web-port when running the instances.

For partitions consumption, im using the topic “greetings_topic” as a source topic to write into the GlobalTable, and every instance is assigned to a different partition of the “greetings_topic”.

The problem is that instances can’t  consume/write data from/into the GlobalTable but in the exact same partition of the source event, as if it is a local “Table”.

[...]

PR #426, opened 2022-12-23
Global global tables

aoberegg:
## Description
If we use_partitioner and global tables it could happen that we write in Worker A to a partitions which is not active in Worker A but active in Worker B. To let Worker B consume the update we have to have all_partitions as standbys as well. 

A similar situation is happening if Global tables are shared over multiple consumer groups by using the possibility to pass a topic that should be used as changelog to the table. Consumer group A could write to the table and consumer group B, C, D only consuming. With the global_global flag it's possible to have shared state over multiple consumer groups.

As a side note. I'm happy if someone has better naming suggestions of the attribute :).


wbarnha:
I like this idea! How about instead we change `is_global_global` to `synchronize_partitions`?

wbarnha:
Potentially related issue? https://github.com/faust-streaming/faust/issues/283

aoberegg:
Hm I'm not quite sure if this fixes https://github.com/faust-streaming/faust/issues/283. Because I think my fix only applies when a different worker writes into the partition of a table that is not his own active. 
Worker A --> actives(1,2), standbys(3,4)
Worker B --> actives(3,4), standbys(1,2)
If Worker A writes into the table partition 3. Then Worker B would not recognize the write because B's assumption is that he is the owner of that partition (as it's his active) and doesn't have to listen to the changelog because he is (assumed) anyway the only one writing to that partition. I'm not quite sure if a scenario like this is happening in https://github.com/faust-streaming/faust/issues/283 because they are not using "use_partitioner" and only writing to the partitions they are also receiving. 

For the naming. Maybe "synchronize_active_partitions" would be something declarative?

wbarnha:
> Hm I'm not quite sure if this fixes #283.

Indeed it doesn't unfortunately, I had to read through your code more thoroughly,

> For the naming. Maybe "synchronize_active_partitions" would be something declarative?

Agreed, I'm much happier with that naming scheme. Once the naming is fixed, I'll be happy to merge in these changes.

wbarnha (changes requested):
Requesting changes via this as a formality. Just change all references from `global_global` to `synchronize_active_partitions` or something equivalently and meaningfully named and we'll be golden!

wbarnha (approved):

Issue #435, opened 2023-01-23
ValueError: The future belongs to a different loop than the one specified as the loop argument


dabdada:
## Checklist

- [x] I have included information about relevant versions
- [ ] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Create a fastapi service and run faust on startup in the same loop. Use uvloop in uvicorn

## Expected behavior

Services should start without issues

## Actual behavior

Some services like Producer raise an error that they are attached to a different loop.

## Full traceback

```pytb
future: <Task finished name='Task-4' coro=<Service.start() done, defined at .../lib/python3.10/site-packages/mode/services.py:799> exception=AssertionError('Please create objects with the same loop as running with')>   call_trace=/Users/gellertd/.pyenv/versions/3.10.6/lib/python3.10/asyncio/runners.py L44
Traceback (most recent call last):
  File .../lib/python3.10/site-packages/mode/services.py", line 800, in start
    await self._default_start()
  File ".../lib/python3.10/site-packages/mode/services.py", line 807, in _default_start
    await self._actually_start()
  File ".../lib/python3.10/site-packages/mode/services.py", line 831, in _actually_start
    await child.maybe_start()
  File ".../lib/python3.10/site-packages/mode/services.py", line 859, in maybe_start
    await self.start()
  File ".../lib/python3.10/site-packages/mode/services.py", line 800, in start
    await self._default_start()
  File ".../lib/python3.10/site-packages/mode/services.py", line 807, in _default_start
    await self._actually_start()
  File ".../lib/python3.10/site-packages/mode/services.py", line 824, in _actually_start
    await self.on_start()
  File ".../lib/python3.10/site-packages/faust/transport/drivers/aiokafka.py", line 1301, in on_start
    await producer.start()
  File ".../lib/python3.10/site-packages/aiokafka/producer/producer.py", line 345, in start
    assert (
AssertionError: Please create objects with the same loop as running with```

# Versions

* Python version: 3.10.6
* Faust version: >=0.10.0
* Operating system: MacOS 13.1 (22C65)

PR #441, opened 2023-02-01
fix: set loop for services where missing

dabdada:
Some services do still start a different / new loop or do not pass the loop param to their super Service call. This brings issues when using faust in other frameworks and using third party loops, e.g. fastapi + uvloop

Sorry I did miss some in the last merge request, but I think I did catch all now. Feel free to check twice 😊 

Fixes https://github.com/faust-streaming/faust/issues/435

### DIFF
--- a/extra/tools/verify_doc_defaults.py
+++ b/extra/tools/verify_doc_defaults.py
@@ -15,36 +15,36 @@ import mode
 from yarl import URL
 
 
-SETTINGS: Path = Path('docs/userguide/settings.rst')
+SETTINGS: Path = Path("docs/userguide/settings.rst")
 
-app = faust.App('verify_defaults')
+app = faust.App("verify_defaults")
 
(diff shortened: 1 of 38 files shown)


wbarnha:
Looks like your PR needs to be linted.

I turned off linting warnings in pre-commits because they got annoying with the false negatives they raised. Maybe I should look into that again so that users don't need to worry about it 😓

dabdada:
> Looks like your PR needs to be linted.
> 
> I turned off linting warnings in pre-commits because they got annoying with the false negatives they raised. Maybe I should look into that again so that users don't need to worry about it 😓

Ok, sorry have hoped for pre-commit to handle this :D

wbarnha:
Looks like there was a lot more linting than I expected 😅 but it looks good to me! LGTM.

wbarnha (approved):

PR #452, opened 2023-02-22
Update pip installs from remaining faust to faust-streaming

gpapia:
Replace `pip install faust` by `pip install faust-streaming` inside the documentation (mainly) and one remaining inside the README.


wbarnha:
Thanks for the PR! I haven't been able to maintain this project as actively as I used to, so all help is greatly appreciated.

I need to re-lint the project before merging this.

wbarnha (approved):

PR #470, opened 2023-03-16
Add support for rocksdict

### LINKED ISSUE #469: Migrate from python-rocksdb to rocksdict
We should give users the option to install `rockdict` as an alternative to `faust-streaming-rocksdb` for a few reasons:

1. I have profiled the performance of `rocksdict` personally and have noticed a 20x boost in performance. There's also an abundance of information on https://github.com/Congyuwang/RocksDict demonstrating its performant abilities
2. `rocksdict` is actively maintained, I have not been able to actively support `faust-streaming-rocksdb` as much
3. `rocksdict` wheels are only 4 MB, whereas `faust-streaming-rocksdb` wheels are 66 MB due to its intensive CI/CD build. This is undesirable because it can increase dependency installation time and Docker image size.

I think adding `faust-streaming[rocksdict]` may be the right way to approach this for now until we know beyond a reasonable doubt that we can phase out `faust-streaming-rocksdb`.

wbarnha:
This will close #469 when it's done. After removing `faust-streaming-rocksdb` from the Docker image, I saw a 250 MB reduction in size.

### DIFF
--- a/faust/stores/rocksdb.py
+++ b/faust/stores/rocksdb.py
@@ -26,7 +26,6 @@ from typing import (
     cast,
 )
 
-import pkg_resources
 from mode.utils.collections import LRUCache
 from yarl import URL
 
@@ -65,12 +64,11 @@ else:
 
 
 try:  # pragma: no cover
-    # Check if rocksdict is installed before using python-rocksdb
-    pkg_resources.get_distribution("rocksdict")
-    ROCKSDICT_INSTALLED = True
     import rocksdict
     from rocksdict import Options, Rdict as DB, WriteBatch  # noqa F811
-except pkg_resources.DistributionNotFound:  # pragma: no cover
+
+    ROCKSDICT_INSTALLED = True
+except ImportError:  # pragma: no cover
     ROCKSDICT_INSTALLED = False
     rocksdict = None  # noqa
 
--- a/faust/stores/rocksdb.py
+++ b/faust/stores/rocksdb.py
@@ -69,7 +69,7 @@ try:  # pragma: no cover
     pkg_resources.get_distribution("rocksdict")
     ROCKSDICT_INSTALLED = True
     import rocksdict
-    from rocksdict import Options, Rdict as DB, WriteBatch
+    from rocksdict import Options, Rdict as DB, WriteBatch  # noqa F811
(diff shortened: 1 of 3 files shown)


wbarnha:
I've been testing this via:
```bash
faust -A examples.word_count worker -l info
```
Everything so far works fine, so all we need to do is adjust unit tests.

faust/stores/rocksdb.py
```diff
@@ -26,6 +26,7 @@
     cast,
 )
 
+import pkg_resources
```
wbarnha, line 29: On second thought, I really don't need this import. I should probably remove it before merging.

wbarnha:
While this doesn't impact the main functionality of Faust with python-rocksdb, we should still write some unit tests at some point. The problem is that the unittests break when `rocksdict` gets installed alongside `faust-streaming-rocksdb`. However, users can still use Faust with both libraries actually installed.

wbarnha:
I think we should give users an option to switch between `rocksdict` and python-rocksdb.

faust/stores/rocksdb.py
```diff
@@ -116,25 +126,57 @@ def __init__(
...
+        if ROCKSDICT_INSTALLED:
+            db_options = self.as_options()
+            db_options.set_db_paths(
+                [rocksdict.DBPath(str(path), self.target_file_size_base)]
+            )
+            db = DB(str(path), options=self.as_options())
+            db.set_read_options(rocksdict.ReadOptions(raw_mode=True))
+            return db
+        else:
+            return rocksdb.DB(str(path), self.as_options(), read_only=read_only)
```
patkivikram, line 138: Should we reset `ROCKSDICT_INSTALLED = False` here?

wbarnha: I think it's best we set that variable based on its import status:
https://github.com/faust-streaming/faust/pull/470/files#diff-083670c0eb5020882953ec86588aa93894a9c8d1371d5806e1c2da38807b5382R66-R75

However, if we add an option for users to selectively switch between the two, then that option would change the state of `ROCKSDICT_INSTALLED`.

wbarnha: Took care of the feature in #470, let's continue there.

PR #472, opened 2023-03-17
Allow users to specify RocksDB drivers

wbarnha:
Review after merging #470.


wbarnha:
Just tested with the example in `word_count` via:
```python
word_counts = app.Table('word_counts', default=int,
                        help='Keep count of words (str to int).',
                        options={"driver": "python-rocksdb"}
                        )
```
Good to go when specifying `python-rocksdb` or `rocksdict`, LGTM.

PR #473, opened 2023-03-17
Bump click requirement to <9.0

wbarnha:
Saw this in https://github.com/robinhood/faust/pull/773, figured I'd change things here.


wbarnha:
Actually never mind, since 8.1.3 is the latest release, let's hold off.

PR #488, opened 2023-04-01
Ack filtered messages when using .take()

artyomilyin:
The issue:
When using filter().take() chain, the filtered out messages are not getting acked (if `NO_CYTHON=False`).
There has been a PR #208 which aimed to fix the same issue, but it only fixed it for installations that don't use `cython`. For `NO_CYTHON=False` this solution didn't work.

Solution:
Compare the message value to the `skipped_value` and ack it.

Note:
The test I added runs red before fix and green after fix. It may look a little bit clumsy but it shows the issue.
Any suggestions are welcome.

### DIFF
--- a/faust/_cython/streams.pyx
+++ b/faust/_cython/streams.pyx
@@ -109,8 +109,7 @@ cdef class StreamIterator:
             object consumer
         consumer = self.consumer
         last_stream_to_ack = False
-        # if do_ack and event is not None:
-        if event is not None and (do_ack or event.value is self._skipped_value):
+        if do_ack and event is not None:
             message = event.message
             if not message.acked:
                 refcount = message.refcount
--- a/faust/streams.py
+++ b/faust/streams.py
@@ -1065,7 +1065,7 @@ class Stream(StreamT[T_co], Service):
                         yield value
                 finally:
                     event, self.current_event = self.current_event, None
-                    it.after(event, do_ack, sensor_state)
+                    it.after(event, do_ack or value is skipped_value, sensor_state)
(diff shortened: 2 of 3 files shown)


wbarnha:
This is awesome, thank you so much for the PR!

faust/_cython/streams.pyx
```diff
@@ -109,8 +109,7 @@ cdef class StreamIterator:
             object consumer
         consumer = self.consumer
         last_stream_to_ack = False
-        # if do_ack and event is not None:
-        if event is not None and (do_ack or event.value is self._skipped_value):
+        if do_ack and event is not None:
```
wbarnha, line 112: I want to approve this, but I wish I could understand why `if do_ack and event is not None:` was disabled in https://github.com/faust-streaming/faust/pull/208. 

But, since you add `or value is skipped_value` below and pass this into this function, these changes are safe to merge.

wbarnha (approved):

faust/_cython/streams.pyx
wbarnha: Everything is explained in https://github.com/faust-streaming/faust/pull/208#issuecomment-1488540746, thank you for that.

artyomilyin: I believe the intention was to ack the message anyway if it's skipped, even if `do_ack` is `False`, but this exact case was not properly tested. The mistake here is `event.value is self._skipped_value` will always return `False`. This is so because in this context `event.value` is the original value of the message, it can never be `skipped_value`.
Thanks for the approval!

PR #490, opened 2023-04-03
Change requirements to set 3.8.0 as minimum aiohttp version

### LINKED ISSUE #484: Please make aiohttp optional
May be possible make aiohttp optional? I dont have any views in faust app, so why i need install aiohttp?

wbarnha:
Closes #484.


wbarnha:
The original reason I set 3.8.3 as the minimum because it's the first version to support Python 3.11. We don't necessarily need this to be true...

PR #493, opened 2023-04-12
Align orjson datetime format with default datetime format

ofirt-orca:

## Description

Pull request https://github.com/faust-streaming/faust/pull/294 re-introduced the orjson serialization. The problem is that for datetime objects it is not backward-compatible to the default format that was used.
The default format (without orjson, see on_default function) used the "Z" suffix for UTC, while the default of orjson is to use the "+00:00" format when serializing datetime.
One can argue that orjson can just be removed from dependencies but sometimes orjson exist for other reasons (for example when using "fast" extra), and in these cases, orjson serialization is "forced" even if it is not wanted.

In case this solution is not desired, there is also an option to add a configuration whether to use orjson or not, or maybe to pass the orjson options somehow.

that version series.


wbarnha:
https://pypi.org/project/codecov/ doesn't exist anymore... did something happen?

wbarnha:
https://github.com/codecov/codecov-python was deprecated, I see now ☹️

wbarnha (approved):
Fixed the pipeline, LGTM and deployed!

Issue #496, opened 2023-04-19
Add support for aerospike > 8.0


patkivikram:
## Checklist
Current faust implementation does not support aersopike > 8.0
- [x ] I have included information about relevant versions
- [ x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Tell us what you did to cause something to happen.

## Expected behavior

Tell us what you expected to happen.

## Actual behavior

Tell us what happened instead.

## Full traceback

```pytb
Paste the full traceback (if there is any)
```

# Versions

* Python version
* Faust version 0.10.11
* Operating system
* Kafka version
* RocksDB version (if applicable)

wbarnha:
I believe merging #498 sufficiently allows us to close this now.

PR #498, opened 2023-04-26
updating version to 8.0.0 compatible client

patkivikram:
## Description

This change includes changing the aerospike client calls to support aerospike client > 8.0 as described [here](https://developer.aerospike.com/client/python/incompatible#version-800)
This fixes issue https://github.com/faust-streaming/faust/issues/496


wbarnha:
Always the silly linting checks that get us :smiling_face_with_tear:

wbarnha:
Aha, found one at https://github.com/wearerequired/lint-action, this would be a great opportunity to test it!

PR #510, opened 2023-05-26
Ensure callback map is rebuilt after reply_to topics are created

### LINKED ISSUE #298: agent.ask() not retrieving value

richardhundt:
A startup race on adding topics to the conductor when `app.conf.reply_create_topic == True` and the app is started in client only mode: `app.start_client()` causes a lookup in the callback map in the conductor before the callbacks have been registered.

Fixes #298 

[EDIT]
Actually, I think the callbacks never get built regardless, because the side-effect of setting `_flag_changes` doesn't rebuild `_tp_index` or `_topic_name_index`, so calling `add` to add a topic after `on_client_only_start` is the issue. The application startup is a little hard to reason about, but that's my understanding.

### DIFF
--- a/faust/agents/replies.py
+++ b/faust/agents/replies.py
@@ -174,8 +174,8 @@ class ReplyConsumer(Service):
             self._fetchers[topic_name] = None
             # declare the topic
             topic = self._reply_topic(topic_name)
-            self.app.topics.add(topic)
             await topic.maybe_declare()
+            self.app.topics.add(topic)
             await self.sleep(3.0)
             # then create the future
             self._fetchers[topic_name] = self.add_future(self._drain_replies(topic))
--- a/faust/agents/replies.py
+++ b/faust/agents/replies.py
@@ -174,6 +174,7 @@ class ReplyConsumer(Service):
             self._fetchers[topic_name] = None
             # declare the topic
             topic = self._reply_topic(topic_name)
+            self.app.topics.add(topic)
             await topic.maybe_declare()
             await self.sleep(3.0)
             # then create the future
--- a/faust/transport/conductor.py
+++ b/faust/transport/conductor.py
@@ -338,10 +338,6 @@ class Conductor(ConductorT, Service):
                     self._acking_topics.add(topic)
                 self._topic_name_index[topic].add(channel)
 
-        if self.app.client_only:
-            self._init_tp_index()
-            self._update_callback_map()
-
         return self._topic_name_index
 
     async def on_partitions_assigned(self, assigned: Set[TP]) -> None:
@@ -352,6 +348,7 @@ class Conductor(ConductorT, Service):
         T(self._update_callback_map)()
 
     async def on_client_only_start(self) -> None:
+        self.log.info("Starting client only mode with %s topics...", self._topics)
         self._init_tp_index()
         self._update_callback_map()
 
@@ -420,7 +417,7 @@ class Conductor(ConductorT, Service):
 
     def _topic_contain_unsubscribed_topics(self, topic: TopicT) -> bool:
         index = self._topic_name_index
-        return bool(index and any(t not in index for t in topic.topics))
+        return bool(any(t not in index for t in topic.topics))
 
     def discard(self, topic: Any) -> None:
         """Unregister topic from conductor."""
--- a/faust/transport/conductor.py
+++ b/faust/transport/conductor.py
@@ -348,15 +348,12 @@ class Conductor(ConductorT, Service):
         T(self._update_callback_map)()
 
     async def on_client_only_start(self) -> None:
-        self._init_tp_index()
-        self._update_callback_map()
-
-    def _init_tp_index(self) -> None:
         tp_index = self._tp_index
         for topic in self._topics:
             for subtopic in topic.topics:
                 tp = TP(topic=subtopic, partition=0)
                 tp_index[tp].add(topic)
+        self._update_callback_map()
 
     def _update_tp_index(self, assigned: Set[TP]) -> None:
         assignmap = tp_set_to_map(assigned)
--- a/faust/transport/conductor.py
+++ b/faust/transport/conductor.py
@@ -338,8 +338,10 @@ class Conductor(ConductorT, Service):
                     self._acking_topics.add(topic)
                 self._topic_name_index[topic].add(channel)
 
-        self._init_tp_index()
-        self._update_callback_map()
+        if self.app.client_only:
+            self._init_tp_index()
+            self._update_callback_map()
+
         return self._topic_name_index
 
     async def on_partitions_assigned(self, assigned: Set[TP]) -> None:
--- a/faust/transport/conductor.py
+++ b/faust/transport/conductor.py
@@ -338,6 +338,8 @@ class Conductor(ConductorT, Service):
                     self._acking_topics.add(topic)
                 self._topic_name_index[topic].add(channel)
 
+        self._init_tp_index()
+        self._update_callback_map()
         return self._topic_name_index
 
     async def on_partitions_assigned(self, assigned: Set[TP]) -> None:
@@ -348,12 +350,15 @@ class Conductor(ConductorT, Service):
         T(self._update_callback_map)()
 
     async def on_client_only_start(self) -> None:
+        self._init_tp_index()
+        self._update_callback_map()
+
+    def _init_tp_index(self) -> None:
         tp_index = self._tp_index
         for topic in self._topics:
             for subtopic in topic.topics:
                 tp = TP(topic=subtopic, partition=0)
                 tp_index[tp].add(topic)
-        self._update_callback_map()
 
     def _update_tp_index(self, assigned: Set[TP]) -> None:
         assignmap = tp_set_to_map(assigned)
--- a/faust/transport/conductor.py
+++ b/faust/transport/conductor.py
@@ -348,7 +348,6 @@ class Conductor(ConductorT, Service):
         T(self._update_callback_map)()
 
     async def on_client_only_start(self) -> None:
-        self.log.info("Starting client only mode with %s topics...", self._topics)
         self._init_tp_index()
         self._update_callback_map()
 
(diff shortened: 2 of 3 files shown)


richardhundt:
I'm not 100% clear on whether this is the most elegant way to solve it, but it seems clear that doing the initialization of `_tp_index` and `_tp_to_callback` during `on_client_only_start` is too early. The `add` call for the `reply_to` topic runs when `self._topic_name_index` is the empty `set()` and so `_topic_contain_unsubscribed_topics` is always false (short circuits on `bool(index and any(...))`, so `self._flag_changes()` doesn't trigger.

wbarnha:
Very interesting, thanks for the PR! I'll need a bit to review this.

richardhundt:
Here's a test case. The `test_case.py` is the server, and `test_client.py` snippet is the client. This should let you reproduce.

```python
# test_case.py
import faust
import asyncio

IS_CLIENT = os.environ.get("IS_CLIENT", False)

def create_app():
    kwargs = dict(
        broker="kafka://localhost:9092",
        origin="test_case",
        topic_partitions=1,
        topic_allow_declare=True,
    )
    if IS_CLIENT:
        kwargs["reply_create_topic"] = True
        kwargs["loop"] = asyncio.get_event_loop()
        return faust.App("test_client", **kwargs)
    else:
        return faust.App("test_case", **kwargs)

_faust_app = None
def app_instance() -> faust.App:
    global _faust_app
    if _faust_app is None:
        _faust_app = create_app()
    return _faust_app


app = app_instance()
ping_topic = app.topic(
    "test-ping",
    value_type=str,
    partitions=1
)


@app.agent(ping_topic)
async def ping_agent(stream):
    count = 0
    async for msg in stream:
        print("got message", msg)
        yield f"pong-{count}"

if __name__ == '__main__':
    app.main()
```

```python
# test_client.py
import asyncio
import os
import logging

LOGLEVEL = os.environ.get('LOGLEVEL', 'INFO').upper()
logging.basicConfig(level=LOGLEVEL)

os.environ["IS_CLIENT"] = "1"

loop = asyncio.get_event_loop_policy().new_event_loop()
asyncio.set_event_loop(loop)

async def pinger():
    from test_case import app_instance, ping_agent

    app = app_instance()
    await app.start_client()

    while True:
        print("loop top")
        await asyncio.sleep(3)
        print("sending ping...")
        pong = await ping_agent.ask(key="42", value="ping")
        print("received ", pong)

if __name__ == '__main__':
    loop.run_until_complete(pinger())
```

richardhundt:
I've updated the pull request with an alternative fix which moves the responsibility of ensuring topic registration with the conductor from the `ReplyConsumer`. I think this is more like what you'd want. The previous fix seemed like a bit of a hack.

faust/agents/replies.py
```diff
@@ -174,6 +174,7 @@ async def _start_fetcher(self, topic_name: str) -> None:
             self._fetchers[topic_name] = None
             # declare the topic
             topic = self._reply_topic(topic_name)
+            self.app.topics.add(topic)
```
wbarnha, line 177: I'm surprised this isn't called already. I'll need to double-check things.

faust/agents/replies.py
```diff
@@ -174,6 +174,7 @@ async def _start_fetcher(self, topic_name: str) -> None:
...
             await topic.maybe_declare()
             await self.sleep(3.0)
```
wbarnha, line 179: Not related to your PR directly but thanks for bringing this to my attention- I wonder why `3.0` is a hard-coded value.

wbarnha:
I like the new changes much more! Thank you for reworking things.

faust/agents/replies.py
richardhundt: It's also sensitive to ordering. That needs to happen before `topic.maybe_declare()`, or it still fails.

The reply topic is created lazily, it seems. `Conductor::on_client_only_start` runs early, when there are no topics added, so it fails to build the callback map. When `topic.maybe_declare()` does finally ask the producer to create the topic, the conductor doesn't know anything about it, but it still expects there to be a callback installed.

faust/agents/replies.py
richardhundt: It works in my test without that `sleep`... might be spurious

richardhundt: There's an explicit test for it though :)
```python
       c._drain_replies.assert_called_once_with(topic)
       c.sleep.assert_called_once_with(3.0)
```

faust/transport/conductor.py
```diff
@@ -413,7 +413,7 @@ def add(self, topic: TopicT) -> None:
 
     def _topic_contain_unsubscribed_topics(self, topic: TopicT) -> bool:
         index = self._topic_name_index
-        return bool(index and any(t not in index for t in topic.topics))
+        return bool(any(t not in index for t in topic.topics))
```
wbarnha, line 416: Not sure how I feel about removing the check for `index`, I need to think about edge cases where this could cause things to break.

richardhundt: That literally says: "only check that that the topics are in `index` if `index` is non-empty"... the only case that I can think of where you might want to have that behaviour is if you want to be sure that there's an assignor subscription, but that doesn't make much sense to me as there's surely a more explicit way of doing that?

wbarnha: > the only case that I can think of where you might want to have that behaviour is if you want to be sure that there's an assignor subscription

Good point. I looked through all the code we have and it seems we don't need to worry about that, I was being overly cautious. I also reviewed the git blame and this method hasn't gotten attention since 2018, when the project was undergoing more changes. Looks like Ask included the `index` check to support the `Mock` return_value.

faust/agents/replies.py
wbarnha: I'll need to review how the callbacks are installed. Because if `maybe_declare` fails, then a topic will be erroneously added to the conductor's topic list. It would make more sense to add a topic to the list of conductor topics after successful topic creation like we do with leader topics in https://github.com/faust-streaming/faust/blob/master/faust/assignor/leader_assignor.py#L24-L28. Let me dig into Faust's internals to see if we can install a callback in advance.

richardhundt: It seems that it is in fact *not* sensitive to ordering. My bad. I think hit a Kafka re-balancing, and couldn't tell the difference between the latency and it hanging.

wbarnha:
I apologize for the lack of progress on this PR. I didn't see https://github.com/faust-streaming/faust/pull/510/commits/2bbc0b82348cc5e70854d807197b5f6d643fe64b get committed. I'm comfortable with this PR now.

wbarnha (approved):
I appreciate the `self.app.topics.add(topic)` call being moved to after topic declaration, makes much more sense now. :slightly_smiling_face: 

LGTM!

patkivikram (approved):

PR #518, opened 2023-06-13
Fixes AttributeError: 'list' object has no attribute 'difference' in …

richardhundt:
## Description

The `SetTableManager` is falling over during updates because the `members` being added is obviously not a `set` and `mode/utils/collections.py` just does `cast(Set, ...)` which just lies to the type system

```
[2023-06-13 13:08:25,368] [222966] [ERROR] [^----Agent*: faust.tables.sets.[.]_modify_set]: Crashed reason=AttributeError("'list' object has no attribute 'difference'") 
  File "/home/richard/Work/pinktum-ai/ai-platform/.venv/lib/python3.10/site-packages/faust/agents/agent.py", line 674, in _execute_actor
    await coro
  File "/home/richard/Work/pinktum-ai/ai-platform/.venv/lib/python3.10/site-packages/faust/tables/sets.py", line 250, in _modify_set
    handler(set_key, members)
  File "/home/richard/Work/pinktum-ai/ai-platform/.venv/lib/python3.10/site-packages/faust/tables/sets.py", line 207, in _update
    self.set_table[key].update(members)
  File "/home/richard/Work/pinktum-ai/ai-platform/.venv/lib/python3.10/site-packages/mode/utils/collections.py", line 622, in update
    added=cast(Set, other).difference(self),
AttributeError: 'list' object has no attribute 'difference'
```

### DIFF
--- a/faust/tables/sets.py
+++ b/faust/tables/sets.py
@@ -245,7 +245,7 @@ class SetTableManager(Service, Generic[KT, VT]):
             except ValueError:
                 self.log.exception("Unknown set operation: %r", set_operation.action)
             else:
-                members = [_maybe_model(m) for m in set_operation.members]
+                members = {_maybe_model(m) for m in set_operation.members}
                 handler = actions[action]
                 handler(set_key, members)
 


wbarnha:
I'm surprised and heartbroken that changing a list to a set could cause the tests to fail...

Out of curiosity, how did you produce this result? I've never seen this error before.

richardhundt:
> I'm surprised and heartbroken that changing a list to a set could cause the tests to fail...
> 
> Out of curiosity, how did you produce this result? I've never seen this error before.

Oh, sorry, I forgot to run the tests. However, that was clearly broken since the handler expects to be passed a `set`.

I'm just using a `SetTable`... and there are some serious issues with that. It's all quite broken. For example, this bit in `faust.tables.objects` does a periodic flush based in `_dirty` keys:

```python
    def flush_to_storage(self) -> None:
        """Flush set contents to storage."""
        for key in self._dirty:
            self.storage[key] = self.data[key].as_stored_value()
        self._dirty.clear()

    @Service.task
    async def _periodic_flush(self) -> None:  # pragma: no cover
        async for sleep_time in self.itertimer(2.0, name="SetManager.flush"):
            self.flush_to_storage()
```

However, there's an assertion in `rocksdb._set` which expects `current_event()` to be not `None`. This is clearly broken since the flush runs asynchronously and there's no `current_event()`:

```python
    def _set(self, key: bytes, value: Optional[bytes]) -> None:
         event = current_event()
         assert event is not None # BOOM: event is not available during _periodic_flush
         partition = event.message.partition
         db = self._db_for_partition(partition)
         self._key_index[key] = partition
         db.put(key, value)
```
Clearly this code was still in flux when Ask disconnected from the project.

richardhundt:
I've added another draft PR which has a tentative fix for the flushing issue (it also includes this tiny patch). I've also done a brute-force hack which uses `murmur2` directly to get the partition if all else fails, but it's probably not necessary.

richardhundt:
> I'm surprised and heartbroken that changing a list to a set could cause the tests to fail...
It's not surprising if the tests are testing the current behaviour instead of the correct behaviour... you can't do set operations on lists ;)

richardhundt:
This doesn't fix the issue :(

Needs more investigation.

PR #520, opened 2023-06-14
Fix for background `flush_to_storage` task not having an event active

richardhundt:
## Description

Track the partition during `send_changelog_event` so that the background periodic flushing task can recover it because there's no `current_event` in scope, and rocksdb storage needs to know.


richardhundt:
Re-opening to separate out the SetTable related changes.

PR #521, opened 2023-06-14
Fix set operations in `ChangeloggedSet`

richardhundt:
## Description

Make sure that sets are recovered during `sync_from_storage` which stores a `list`, and pass actual set objects to the underlying `ManagedUserSet` from mode which is depends on actual `SetLike` objects instead of lists.

### DIFF
--- a/faust/tables/sets.py
+++ b/faust/tables/sets.py
@@ -245,7 +245,7 @@ class SetTableManager(Service, Generic[KT, VT]):
             except ValueError:
                 self.log.exception("Unknown set operation: %r", set_operation.action)
             else:
-                members = [_maybe_model(m) for m in set_operation.members]
+                members = {_maybe_model(m) for m in set_operation.members}
                 handler = actions[action]
                 handler(set_key, members)
 
--- a/faust/tables/sets.py
+++ b/faust/tables/sets.py
@@ -84,7 +84,7 @@ class ChangeloggedSet(ChangeloggedObject, ManagedUserSet[VT]):
         self.manager.send_changelog_event(self.key, OPERATION_UPDATE, [added, removed])
 
     def sync_from_storage(self, value: Any) -> None:
-        self.data = cast(Set, value)
+        self.data = set(value)
 
     def as_stored_value(self) -> Any:
         return self.data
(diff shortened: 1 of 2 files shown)


wbarnha:
Looks like everything is passing now. Is this draft ready for review?

richardhundt:
> Looks like everything is passing now. Is this draft ready for review?

I've marked it as ready. Thanks!

faust/tables/sets.py
```diff
@@ -84,7 +84,7 @@ def on_change(self, added: Set[VT], removed: Set[VT]) -> None:
         self.manager.send_changelog_event(self.key, OPERATION_UPDATE, [added, removed])
 
     def sync_from_storage(self, value: Any) -> None:
-        self.data = cast(Set, value)
+        self.data = set(value)
```
wbarnha, line 87: I've always been baffled by the usage of `typing.cast` sprinkled all over, but it could just be remnants of the project from older versions of Python where typing was still a WIP.

wbarnha (commented):
Since lists can be converted to sets without any issue (assuming there are never duplicate entries), this looks like a logical operation to me. Regarding your usage of `sync_from_storage`, I am guessing you've had issues with data recovery from stores into table instances? If you have tracebacks demonstrating the issue, it'd make me feel much more comfortable merging this.

I'd like to test this on some Faust examples just to be sure everything works and I'd also like to defer to @patkivikram to make sure that there's no edge cases I'm forgetting.

richardhundt:
> Since lists can be converted to sets without any issue (assuming there are never duplicate entries), this looks like a logical operation to me. Regarding your usage of `sync_from_storage`, I am guessing you've had issues with data recovery from stores into table instances? If you have tracebacks demonstrating the issue, it'd make me feel much more comfortable merging this.
> 
> I'd like to test this on some Faust examples just to be sure everything works and I'd also like to defer to @patkivikram to make sure that there's no edge cases I'm forgetting.

I don't have stack traces, but It cannot possibly work. If you look at `mode.collections.ManagedUserSet.update`:

```python
    def update(self, other: _Setlike[T]) -> None:
        # union update
        self.on_change(
            added=cast(Set, other).difference(self),  # HERE: this fails because we've passed a list for `other`
            removed=set(),
        )
        cast(Set, self.data).update(other)
```

This is all I'm doing:

```python
account_sessions_table: faust.SetTable = app.SetTable(
    name='account_sessions',
    start_manager=True
)

@app.agent(sessions_topic)
async def session_agent(stream: faust.StreamT[Session]):
    async for session in stream:
        await account_sessions_table.manager.add(key=session.account_id, member=session.id)
        yield session
```

It's a completely trivial use of `SetTable` and it fails because the update is being given a list. Try any simple RocksDB backed use of `SetTable` and it'll fail. This isn't some obscure corner case.

wbarnha (approved):
After further examination and testing, your changes here look good! I forgot how much `Set` gets passed around with improper typing. Your changes have brought attention to the need for fixing other ongoing typing issues. Thank you very much, LGTM.

Issue #497, opened 2023-04-20
Writing into Table with Stream.take() not allowed


dabdada:
## Steps to reproduce

When using the Stream.take() function I can not write into a Table but get the error `Cannot modify table key from outside of stream iteration`.

Can I just not write to tables when using take?

Example:

```
from faust import App

app = App(id="take-error-example")

topic = app.topic('orders')
table = app.Table('orders')

@app.agent(topic)
async def consumer(stream):
    async for values in stream.take(1):
        for value in values:
            table["key"] = "some value"
            yield value
```

## Expected behavior

The key value pair should be written into the Table

## Actual behavior

```pytb
File ".../python3.10/site-packages/mode/utils/collections.py", line 652, in __setitem__
    self.on_key_set(key, value)
  File ".../python3.10/site-packages/faust/tables/table.py", line 77, in on_key_set
    fut = self.send_changelog(self.partition_for_key(key), key, value)
  File ".../python3.10/site-packages/faust/tables/base.py", line 314, in partition_for_key
    raise TypeError(
TypeError: Cannot modify table key from outside of stream iteration
```

# Versions

* Python version: 3.10.6
* Faust version: 0.10.11
* Operating system: MacOS 13

dabdada:
Workaround for now is that we use a sink for this or read another topic

mcskatkat:
> Workaround for now is that we use a sink for this or read another topic

Can you explain the workaround? Modified pseudocode from the example perhaps?
I'm afraid of hitting the same issue and I need `.take(N)` to work for large N as we benefit from batch processing.

dabdada:
@mcskatkat  Sorry after checking the source code more thoroughly a sink approach is possibly not suitable as the current event is checked against the processed event and this must be the same event (which it isn't as current event is the last event in take, hence the error). You might need to add a new channel (if not backed by kafka else a topic) and publish all the data consumed via take to this channel. consume the channel without a take approach to modify the table. Might not work in all situations though but worked for us...

mcskatkat:
Thanks @dabdada .
Actually in my case the table update uses associative reductions on events yielded from `.take()`, so I'm totally ok making just one table update with the "current event" being the last in the batch. Much better than accessing the table for every event.

However from your example I gather that this is impossible, as Faust will raise an exception because I didn't `iter` the `stream` itself.

mcskatkat:
Thanks. I suppose it's worth giving it a try then.

richardhundt:
[...]

Issue #261, opened 2022-01-19
SetTable crashes when using RocksDB


elja:
## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

If you use "rocksdb" as storage you got this error, it doesn't happen if you use memory. 

Basically, I have a list of tasks that I load from the database and store locally (no kafka here), then I have some external events that are triggering execution of these tasks, the only thing I want is to not send a task to a kafka if it's already was sent and currently processing. I use a SetTable to track the progress of the task execution. All topics and this table have the same number of partitions and also use the same key, so they should be co-partitioned. 

```python
app = faust.App(
    id=f"scheduler_{config.ENV}",
    broker=config.KAFKA_BOOTSTRAP_SERVERS,
    store='rocksdb://'
)

events_topic = app.topic('external_events', partitions=8) # events from another service, pertiotioned by some event_key

task_queue_topic = app.topic('task_queue_topic', partitions=8)
task_completion_topic = app.topic('task_completion_topic', partitions=8)
task_progress = app.SetTable('task_progress', partitions=8, value_type=int)

async def task_queue_sink(task):
    logging.info(f"Sending task: task_id = {task.task_id} / uuid = {task.uuid}")
    await task_queue_topic.send(key=task.event_key, value=task)

@app.agent(events_topic, sink=[task_queue_sink])
async def events_stream(stream: faust.Stream):
    async for key, event in stream.items():
        tasks = await maybe_load_tasks_from_db(event_key=key) # or event.event_key

        for task in tasks:
            if task.task_id not in task_progress[event_key]:
                task_progress[event_key].add(task.task_id)
                yield task

@app.agent(task_completion_topic)
async def task_completion_stream(stream: faust.Stream):
    async for task in stream:
        logging.info(f"Task completed: task_id = {task.task_id} / uuid = {task.uuid}")
        task_progress[task.event_key].discard(task.task_id)
```

Tell us what you did to cause something to happen.

## Expected behavior

I should be able to store data to the SetTable that uses RocksDB without any errors.

## Actual behavior

`AssertionError: assert event is not None`
../python3.9/site-packages/faust/stores/rocksdb.py#274

## Full traceback

```pytb



[2022-01-19 14:26:10,389] [33567] [ERROR] [^----ChangeloggedSetManager: None]: Crashed reason=AssertionError() 
Traceback (most recent call last):
[...]

PR #522, opened 2023-06-14
Fix periodic flush not having a `current_event` in context

richardhundt:
## Description

~~Track the partition during send_changelog_event so that the background periodic flushing task can recover it~~ Remove the periodic flush because there's no current_event in scope, and rocksdb storage needs to know.

The periodic flush doesn't seem to add any value since any disk I/O will still be synchronous even if running in a separate coroutine.

### DIFF
--- a/faust/tables/objects.py
+++ b/faust/tables/objects.py
@@ -65,20 +65,18 @@ class ChangeloggedObjectManager(Store):
     data: MutableMapping
 
     _storage: Optional[StoreT] = None
-    _dirty: Set
 
     def __init__(self, table: Table, **kwargs: Any) -> None:
         self.table = table
         self.table_name = self.table.name
         self.data = {}
-        self._dirty = set()
         Service.__init__(self, loop=table.loop, **kwargs)
 
     def send_changelog_event(self, key: Any, operation: int, value: Any) -> None:
         """Send changelog event to the tables changelog topic."""
         event = current_event()
-        self._dirty.add(key)
         self.table._send_changelog(event, (operation, key), value)
+        self.storage[key] = self[key].as_stored_value()
 
     def __getitem__(self, key: Any) -> ChangeloggedObject:
         if key in self.data:
@@ -99,10 +97,6 @@ class ChangeloggedObjectManager(Store):
     async def on_start(self) -> None:
         """Call when the changelogged object manager starts."""
         await self.add_runtime_dependency(self.storage)
-
-    async def on_stop(self) -> None:
-        """Call when the changelogged object manager stops."""
-        self.flush_to_storage()
 
     def persisted_offset(self, tp: TP) -> Optional[int]:
         """Get the last persisted offset for changelog topic partition."""
@@ -132,17 +126,6 @@ class ChangeloggedObjectManager(Store):
         """Sync set contents from storage."""
         for key, value in self.storage.items():
             self[key].sync_from_storage(value)
-
-    def flush_to_storage(self) -> None:
-        """Flush set contents to storage."""
-        for key in self._dirty:
-            self.storage[key] = self.data[key].as_stored_value()
-        self._dirty.clear()
-
-    @Service.task
-    async def _periodic_flush(self) -> None:  # pragma: no cover
-        async for sleep_time in self.itertimer(2.0, name="SetManager.flush"):
-            self.flush_to_storage()
 
     def reset_state(self) -> None:
         """Reset table local state."""
--- a/tests/unit/tables/test_objects.py
+++ b/tests/unit/tables/test_objects.py
@@ -54,6 +54,7 @@ class Test_ChangeloggedObjectManager:
     def man(self, *, table):
         man = ChangeloggedObjectManager(table)
         man.ValueType = ValueType
+        man.storage.__setitem__ = Mock()
         return man
 
     @pytest.fixture()
@@ -62,7 +63,7 @@ class Test_ChangeloggedObjectManager:
 
     def test_send_changelog_event(self, *, man, table, key, current_event):
         man.send_changelog_event(key, 3, "value")
-        assert key in man._dirty
+        assert man.storage.__setitem__.called_once_with(key, "value")
         table._send_changelog.assert_called_once_with(
             current_event(),
             (3, key),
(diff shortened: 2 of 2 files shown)


richardhundt:
I can't grok why codecov shows lower coverage. I deleted stuff mostly.

wbarnha:
I'm not sure either. In `aiokafka`, I changed a sentence in documentation and codecov marked it as a failure as well. Let me see what's going on...

wbarnha:
> The periodic flush doesn't seem to add any value since any disk I/O will still be synchronous even if running in a separate coroutine.

Very interesting. I'm going to try to look through old logs to see what the reasons were for including `flush_to_storage`, and related methods. The documentation in https://faust-streaming.github.io/faust/reference/faust.tables.objects.html#faust.tables.objects.ChangeloggedObjectManager.flush_to_storage sadly tells me nothing. I assume you've tested this with the RocksDB and memory stores?

Speaking of which, I should probably split `rocksdb` into two separate files based on drivers. The current implementation was hastily put together while trying to compromise on retaining as much backwards compatibility possible.

richardhundt:
> > The periodic flush doesn't seem to add any value since any disk I/O will still be synchronous even if running in a separate coroutine.
> 
> Very interesting. I'm going to try to look through old logs to see what the reasons were for including `flush_to_storage`, and related methods. The documentation in https://faust-streaming.github.io/faust/reference/faust.tables.objects.html#faust.tables.objects.ChangeloggedObjectManager.flush_to_storage sadly tells me nothing. I assume you've tested this with the RocksDB and memory stores?
> 
> Speaking of which, I should probably split `rocksdb` into two separate files based on drivers. The current implementation was hastily put together while trying to compromise on retaining as much backwards compatibility possible.

I noticed the rocksdict support interleaved in there. I'm more concerned about the dependency on `current_event`. It would probably be more robust if you make the key a requirement and standardize on the `murmur2` hashing function used by the default kafka partitioner. There's a related issue https://github.com/faust-streaming/faust/issues/497 which shows how fragile depending on having a message in flight is.

faust/tables/objects.py
```diff
@@ -65,20 +65,18 @@ class ChangeloggedObjectManager(Store):
     data: MutableMapping
 
     _storage: Optional[StoreT] = None
-    _dirty: Set
```
wbarnha, line 68: Wow, this `_dirty` attr existed since the very first commit of this file. Looks like it never got replaced by anything better. I'm looking forward to this PR changing that.

wbarnha:
I need to mull over this PR a bit longer. The changes you propose make sense. I'm just trying to figure out what Ask Solem intended with the usage of [`self.itertimer`](https://faust-streaming.github.io/mode/reference/mode.services.html?highlight=itertimer#mode.services.Service.itertimer) in https://github.com/faust-streaming/faust/blob/master/faust/tables/objects.py#L142-L145. 

I also would like to thank you for spurring ongoing discussion in your aforementioned issue. It's been useful for corroborating my personal observations.

Once I get this PR merged along with my own updates to Ci/CD workflows, I will release v0.11.0 with all of your changes. 😄

wbarnha:
I think the changes in this PR could lead to us fixing https://github.com/faust-streaming/faust/issues/261, since the lack of an event caused the error.

As for prior documentation, the only notes I could find are in https://github.com/robinhood/faust/pull/215#issue-380761658:

> Sets are kept in memory for fast operation, and this way we also avoid the overhead of constantly serializing/deserializing the data to RocksDB. Instead we periodically flush changes to RocksDB, and populate the sets from disk at worker startup/table recovery.

It seems the motivation for having a periodic flush stems from the desire to reduce disk I/O, within the context of using a SetTable. After examining references to all classes defined in `sets.py`, it's safe to conclude that these changes will only impact classes defined in `sets.py` and their derivatives defined by users outside of Faust. The last question I'd like to answer in my mind is: "Why were periodic flushes defined for SetTable and none of the other tables?" It makes sense to reduce disk I/O and serialization calls throughout the entire application. 

Maybe I'm mulling for no reason, it could just be that Ask just forgot to update SetTable.

richardhundt:
> I think the changes in this PR could lead to us fixing #261, since the lack of an event caused the error.
> 
> As for prior documentation, the only notes I could find are in [robinhood/faust#215 (comment)](https://github.com/robinhood/faust/pull/215#issue-380761658):
> 
> > Sets are kept in memory for fast operation, and this way we also avoid the overhead of constantly serializing/deserializing the data to RocksDB. Instead we periodically flush changes to RocksDB, and populate the sets from disk at worker startup/table recovery.
> 
> It seems the motivation for having a periodic flush stems from the desire to reduce disk I/O, within the context of using a SetTable. After examining references to all classes defined in `sets.py`, it's safe to conclude that these changes will only impact classes defined in `sets.py` and their derivatives defined by users outside of Faust. The last question I'd like to answer in my mind is: "Why were periodic flushes only defined for SetTable and none of the other tables?" It makes sense to reduce disk I/O and serialization calls throughout the entire application.
> 
> Maybe I'm mulling for no reason, it could just be that Ask just forgot to update SetTable.

I would check whether you have any control over disk I/O in the first place because I'm pretty sure rocksdb would be using mmap. Moreover, in the end you're still doing the same number of write operations, you're just collecting them into batches. I don't see what that buys you.

wbarnha:
> I would check whether you have any control over disk I/O in the first place

Good point. We definitely don't. Our legacy python-rocksdb drivers are all synchronous and barebones to the point where all calls to `_set` in https://github.com/faust-streaming/faust/blob/master/faust/stores/rocksdb.py#L454-L460 attempt to immediately write to the DB object.

Funny enough, that's also where `current_event` is. https://github.com/robinhood/faust/issues seems to be a treasure trove when trying to understand the motives behind @ask's design choices. However, searching in https://github.com/robinhood/faust/issues?q=current_event+ does not tell me as much as I'd hoped.

> Regarding the serialization: I think this is the crux. The thinking it seems is that the usage of SetTable would be bursty, where lots of entries are written by the client at once, and you don't want to keep serializing the entire underlying set for each set operation. This would make more sense than trying to save I/O because RocksDB is heavily write optimized already.

Agreed. The definition of `_dirty` must've been created so that only necessary updates would be getting flushed to storage, instead of the entire table contained in `self.data`.

wbarnha (approved):
After thorough review, I'm ready to approve. LGTM!

Issue #519, opened 2023-06-13
Add support for OAuthbearer auth mechanism


cbcoutinho:
## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

OAuth 2.0 support would allow faust streaming apps to be used in Confluent Cloud clusters configured with custom identity pools.

For comparison, the Confluent python library supports OAuth 2 by providing a callback to the consumer/producer `__init__`.

https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#kafka-client-configuration

## Expected behavior

It would be great to be able to utilize Faust with clusters that require integration with external identity providers.

wbarnha:
We use `aiokafka` but it seems to be possible to use OAuth from the existing issue in https://github.com/aio-libs/aiokafka/issues/691.

wbarnha:
We can close this now!

PR #525, opened 2023-06-24
Implement OAuthbearer auth mechanism

cbcoutinho:

## Description

This PR adds an `OAuthCredentials` class to facilitate the OAuth2 SASL mechanism for producers/consumers. I was able to test this by creating an instance of aiokafka's `AbstractTokenProvider` and generating a token from my IdP inside a Faust worker 

Addresses #519

### DIFF
--- a/faust/auth.py
+++ b/faust/auth.py
@@ -9,6 +9,7 @@ from faust.types.auth import AuthProtocol, CredentialsT, SASLMechanism
 __all__ = [
     "Credentials",
     "SASLCredentials",
+    "OAuthCredentials",
     "GSSAPICredentials",
     "SSLCredentials",
 ]
--- a/faust/auth.py
+++ b/faust/auth.py
@@ -1,6 +1,6 @@
 """Authentication Credentials."""
 import ssl
-from typing import Any, Optional, Union, Tuple, Callable
+from typing import Any, Callable, Optional, Tuple, Union
 
 from faust.types.auth import AuthProtocol, CredentialsT, SASLMechanism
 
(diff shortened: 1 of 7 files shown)


wbarnha:
Argh, needs to be linted. Try running `./scripts/lint` and committing? 

I'm contemplating the merits of enabling that pre-commit hook again to avoid this situation but I don't want to confuse users where it dissuades them from filing PRs.

faust/auth.py
```diff
@@ -49,6 +52,33 @@ def __repr__(self) -> str:
...
+    mechanism: SASLMechanism = SASLMechanism.OAUTHBEARER
+
+    ssl_context: Optional[ssl.SSLContext]
+
+    def __init__(
+        self,
+        *,
+        oauth_cb: AbstractTokenProvider,
+        ssl_context: Optional[ssl.SSLContext] = None,
+    ):
```
wbarnha, line 68: I think `client_id` was supposed to be included in here, based on your test.

wbarnha:
I won't torture you further. If you don't want to write the last test case, I can certainly give it a try on Monday. Thank you for the PR!

wbarnha:
I'd be grateful if you could provide documentation or an example on using OAuth2 SASL.

faust/auth.py
cbcoutinho: Thanks for catching this

cbcoutinho:
> I'd be grateful if you could provide documentation or an example on using OAuth2 SASL with Faust.

Hey @wbarnha I'd be happy to add some docs on how to use the class. Where would be the most logical spot to add some documentation on this feature?

wbarnha (approved):
Just checked the latest commit, thanks a bunch. I was checking for that page earlier but couldn't remember where it was.

LGTM!

wbarnha:
I'm hoping I'll have this released shortly. There are two other PRs I need to wrap up reviewing.

PR #530, opened 2023-06-29
Enable PyPy testing

wbarnha:


wbarnha:
Oddly enough, all tests pass when I run them locally on my machine. Is a dependency somewhere misbehaving?

wbarnha:
LGTM for now, we can patch it further later on.

PR #538, opened 2023-07-26
Use rocksdict version >= v0.3.11

GodTamIt:
## Description

`v0.3.11` to `v0.3.13` contains mostly minor changes but the ones that'll affect faust are:

- https://github.com/Congyuwang/RocksDict/pull/68
- https://github.com/Congyuwang/RocksDict/pull/67

### DIFF
--- a/requirements/extras/rocksdict.txt
+++ b/requirements/extras/rocksdict.txt
@@ -1,1 +1,1 @@
-rocksdict==0.3.13
+rocksdict>=0.3.11,<4.0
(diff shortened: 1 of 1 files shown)


wbarnha:
I initially had `rocksdict` configured to install a pinned version in case of incompatible changes, let's use `>=3.11` from hereon.

wbarnha (changes requested):
Replace with `rocksdict>=0.3.11`

wbarnha (approved):
Thanks, LGTM!

PR #540, opened 2023-08-09
Update python-package.yml to test for Python 3.12

wbarnha:
Python 3.12.0rc1 is out now, might as well see what happens.


wbarnha:
Eh, might as well merge it, it's a dependency problem.

Issue #560, opened 2023-10-27
(docs) Autodetect for schemas


redb0:
Hi everyone!

I recently started using this library

In documentation ([link](https://faust-streaming.github.io/faust/userguide/models.html#schemas)) in example for class `Autodetect`  variable `value_type` is not defined before this line

```python
def loads_value(self, app, message, *, loads=None, serializer=None):
    if loads is None:
        loads = app.serializers.loads_value
    # try to get key_type and serializer from Kafka headers
    headers = dict(message.headers)
    value_type_name = headers.get('ValueType')
    serializer = serializer or headers.get('ValueSerializer')
    if value_type_name:
        value_type = registry[value_type]   # <-- value_type is not defined
        return loads(value_type, message.key, serializer=serializer)  # <-- maybe message.value?
    else:
        return super().loads_value(app, message, loads=loads, serializer=serializer)
```

Maybe it was meant `value_type_name`?

Should I pass `message.key` in the last line? Maybe you meant `message.value`?

And also the comment is a victim of copy-paste: `try to get key_type and serializer from Kafka headers` -> `try to get value_type and serializer from Kafka headers`

IbrahimBagalwa:
Hello @redb0, should the `value_type` variable be replaced with `value_type_name`, `message.key` with `message.value`, and the comment `# try to get key_type and serializer from Kafka headers` with `# try to get value_type and serializer from Kafka headers` in this code snippet?

IbrahimBagalwa:
@redb0, If that's the case, I would like to work on it.

redb0:
@IbrahimBagalwa, yes, you wrote everything correctly

I don't mind if you do it

IbrahimBagalwa:
Thanks @redb0

PR #563, opened 2023-11-07
fix: docs Autodetect for schemas

IbrahimBagalwa:
(Fixes [#560](https://github.com/faust-streaming/faust/issues/560))

## Description
I noticed a discrepancy in the documentation, specifically in the example for the Autodetect class. Here are the details of the issue I fixed:

In the documentation example, the variable value_type was not defined before being referenced in the following lines of code:
```python
value_type = registry[value_type]   # <-- value_type is not defined
return loads(value_type, message.key, serializer=serializer)  # <-- maybe message.value?
```
It appears that there might have been a typo or misunderstanding, and it should have been `value_type_name` instead of `value_type`.

Additionally, I noticed a comment in the code that seemed to be a victim of a copy-paste error. It read: `try to get key_type and serializer from Kafka headers,` which I believe should have been: `try to get value_type and serializer from Kafka headers.`

I fixed the issue by making the necessary changes, and I also raised questions about whether it is intended to pass `message.key` in the last line, and if `message.value` should be considered instead.

### DIFF
--- a/docs/userguide/models.rst
+++ b/docs/userguide/models.rst
@@ -186,13 +186,13 @@ a schema that reads message key and value type from Kafka headers:
                         serializer=None):
             if loads is None:
                 loads = app.serializers.loads_value
-            # try to get key_type and serializer from Kafka headers
+            # try to get value_type and serializer from Kafka headers
             headers = dict(message.headers)
             value_type_name = headers.get('ValueType')
             serializer = serializer or headers.get('ValueSerializer')
             if value_type_name:
-                value_type = registry[value_type]
-                return loads(value_type, message.key,
+                value_type = registry[value_type_name]
+                return loads(value_type, message.value,
                        serializer=serializer)
             else:
                 return super().loads_value(


wbarnha (approved):
Thank you for catching this! I'm surprised such an error has gone unnoticed for so long.

wbarnha:
> (Fixes [#560](https://github.com/faust-streaming/faust/issues/560))
> ## Description
> 
> I noticed a discrepancy in the documentation, specifically in the example for the Autodetect class. Here are the details of the issue I fixed:
> 
> In the documentation example, the variable value_type was not defined before being referenced in the following lines of code:
> 
> ```python
> value_type = registry[value_type]   # <-- value_type is not defined
> return loads(value_type, message.key, serializer=serializer)  # <-- maybe message.value?
> ```
> 
> It appears that there might have been a typo or misunderstanding, and it should have been `value_type_name` instead of `value_type`.
> 
> Additionally, I noticed a comment in the code that seemed to be a victim of a copy-paste error. It read: `try to get key_type and serializer from Kafka headers,` which I believe should have been: `try to get value_type and serializer from Kafka headers.`
> 
> I fixed the issue by making the necessary changes, and I also raised questions about whether it is intended to pass `message.key` in the last line, and if `message.value` should be considered instead.

It seems the `loads` function is left undefined. Faust also defines its own `loads` method which behaves differently if `orjson` is installed, which is different from the regular `loads` defined in https://docs.python.org/3/library/json.html.
https://github.com/faust-streaming/faust/blob/3fb3180c38554aa883f886444fc50d23f5437fbc/faust/utils/json.py#L166-L203

Looking at both of these, it's not immediately apparent to me as to which implementation Ask Solem intended to use in the example above, since there's no `import` declared for it. :frowning: I'll stare at it a bit longer.

To be honest I should be testing this myself to make sure it works. I've never used autodetection before because I've always believed that it to be in best practice to use one schema per Kafka topic, because non-Faust applications may also need to consume from a particular topic and my experiences with https://marcosschroh.github.io/dataclasses-avroschema/faust_records/ show me that, despite how powerful these Records are, they can also make your life quite difficult.

PR #567, opened 2023-11-16
Add a TTL for RocksDB store

### LINKED ISSUE #515: Table TTL
Hi All,

This is more of a question than an issue  - not sure if you have a community channel instead?

i see someone posted on the original faust repo, asking if and how we can expire table entries using TTL in 2018 - https://github.com/robinhood/faust/issues/230 . Has this feature been implemented?  This would be handy to prevent table bloat over time, when not using windowed tables.

YiuRULE:
## Description

Add TTL for faust stores.
Fixes #515

### DIFF
--- a/faust/stores/rocksdb.py
+++ b/faust/stores/rocksdb.py
@@ -113,6 +113,7 @@ class RocksDBOptions:
         block_cache_compressed_size: Optional[int] = None,
         bloom_filter_size: Optional[int] = None,
         use_rocksdict: Optional[bool] = None,
+        ttl: Optional[int] = None,
         **kwargs: Any,
     ) -> None:
         if max_open_files is not None:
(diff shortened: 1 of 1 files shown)


wbarnha (commented):
I'm definitely for adding this feature but I'm wondering if we should allow more options to be specified while we're at it. Could you so provide an example in the docstring as to how this should be utilized?

wbarnha:
Cheers, thanks for the example! I'll review it further when I get to a proper computer in the morning before merging.

YiuRULE:
I add some documentation on the latest revision, let me know if it's ok :)
And also if you wish to add something more if needed to

wbarnha (approved):
Good stuff, thanks for the PR!

PR #569, opened 2023-11-21
Fix potential NoneType access in aiokafka driver

GodTamIt:
## Description

This logic is likely meant to be a boolean `OR` not `AND`, since the first check in the statement checks for `assignment` to be `None`.

### DIFF
--- a/faust/transport/drivers/aiokafka.py
+++ b/faust/transport/drivers/aiokafka.py
@@ -836,7 +836,7 @@ class AIOKafkaConsumerThread(ConsumerThread):
         secs_since_started = now - self.time_started
         aiotp = TopicPartition(tp.topic, tp.partition)
         assignment = consumer._fetcher._subscriptions.subscription.assignment
-        if not assignment and not assignment.active:
+        if not assignment or not assignment.active:
             self.log.error(f"No active partitions for {tp}")
             return True
         poll_at = None


wbarnha (approved):
Genuinely surprised how such a simple thing got overlooked for so long. Makes sense to change it to this.

What caused you to stumble upon this, out of curiosity @GodTamIt?

wbarnha:
Ah crud... looks like I need to go clean up the pyproject.toml file. I promised myself I'd finish switching things over eventually, but now my hand is forced.

GodTamIt:
> Genuinely surprised how such a simple thing got overlooked for so long. Makes sense to change it to this.
> 
> What caused you to stumble upon this, out of curiosity @GodTamIt?

@wbarnha I was running the test in https://github.com/faust-streaming/faust/pull/568 for long enough that it hit this edge case.

It was quite hard to reproduce actually.

wbarnha:
> > Genuinely surprised how such a simple thing got overlooked for so long. Makes sense to change it to this.
> > 
> > What caused you to stumble upon this, out of curiosity @GodTamIt?
> 
> @wbarnha I was running the test in https://github.com/faust-streaming/faust/pull/568 for long enough that it hit this edge case.
> 
> It was quite hard to reproduce actually.

Thank you for finding this! I'll merge it ASAP.

Issue #570, opened 2023-11-21
Unable to install using Docker on Mac.


hgalytoby:
## Steps to reproduce
```dockerfile
FROM python:3.10.13
RUN pip install --upgrade pip
RUN pip3 install faust-streaming
```

```dockerfile
FROM ubuntu:latest
RUN apt-get update \
    && apt-get install -y python3-pip \
    && python3 -m pip install --upgrade pip
RUN pip3 install faust-streaming
```

## Actual behavior

On 2023/11/20, I successfully installed and ran faust-streaming using Docker. However, as of today, 2023/11/21, I am unable to install faust-streaming anymore. I attempted to install faust-streaming separately using two different images, namely Ubuntu and Python 3.10.13, but without success.

I can install faust-streaming locally on my Mac.

I can install faust-streaming on Ubuntu using Docker and a Dockerfile.

I can install faust-streaming on Windows.

I am unsure why using Docker on Mac is presenting this issue. Has anyone else encountered a similar problem?

I conducted separate tests on MacOs M1 versions 14.1.1, 12.6.2, and 13.4.1.

## Full traceback
```
...move
#0 7.735               ********************************************************************************
#0 7.735               The following seems to be defined outside of `pyproject.toml`:
#0 7.735       
[...]

PR #571, opened 2023-11-21
Temporarily fix breaking build with setuptools v69

somnam:

## Description
Temporarily fix breaking build with setuptools v69

https://github.com/faust-streaming/faust/issues/570

that version series.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ dynamic = ["version"]
 
 [build-system]
 requires = [
-    "setuptools",
+    "setuptools>=30.3.0,<69.0.0",
     "setuptools_scm[toml]",
     "wheel",
     "cython>=0.29; implementation_name == 'cpython'",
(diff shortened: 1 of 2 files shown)


wbarnha (approved):
Thanks for the PR, LGTM while I figure out what's going on with this.

somnam:
Thanks for merging! Will there be a new release with this fix?

wbarnha:
I'm in an awkward spot right now with getting a 3.12 release out. Is it blocking you from installing Faust?

somnam:
> I'm in an awkward spot right now with getting a 3.12 release out. Is it blocking you from installing Faust?

On M1 Macs or aarch images it's kind of impossible now due to this bug.
I've worked around it by doing a local wheel build and installing that when on arm/aarch architecture, but it's kind of hacky.

wbarnha:
My condolences, let's get that release out for you.

PR #581, opened 2023-11-30
Add FastAPI example

wbarnha:
There's been an increase in demand for this and I've personally been meaning to add this for some time.


chanpl:
Thanks @wbarnha for providing the example. Can you also please provide an example for medium/large project? In my case, consumers are not subscribing any topics if agent is defined in a separate file.

hgalytoby:
> Thanks @wbarnha for providing the example. Can you also please provide an example for medium/large project? In my case, consumers are not subscribing any topics if agent is defined in a separate file.

For reference?

https://github.com/faust-streaming/faust/pull/595

hgalytoby:
> There's been an increase in demand for this and I've personally been meaning to add this for some time.

@wbarnha 

I've added a new FastAPI example, could you take a look?

wbarnha:
Sure! Thanks for adding it!

chanpl:
@wbarnha Thank you very much! I really appreciate it.

PR #587, opened 2023-12-08
Add official support for Python 3.12

wbarnha:
Let's wrap this up.


AntonOfTheWoods:
Awesome @wbarnha , thanks! Any idea on when there will be a release?

PR #600, opened 2024-01-10
enable kafka static-partition-assignment (KIP-345)

Mopsgeschwindigkeit:
With version v0.10.0 aiokafka implemented and tested KIP-345 (static partition assignment, PR ->https://github.com/aio-libs/aiokafka/pull/941 ).

 With this change the class `aiokafka.AIOKafkaConsumer()` accepts the `group_instance_Id` parameter.

### DIFF
--- a/tests/unit/transport/drivers/test_aiokafka.py
+++ b/tests/unit/transport/drivers/test_aiokafka.py
@@ -787,7 +787,7 @@ class Test_AIOKafkaConsumerThread(AIOKafkaConsumerThreadFixtures):
                 api_version=app.conf.consumer_api_version,
                 client_id=conf.broker_client_id,
                 group_id=conf.id,
-                # group_instance_id=conf.consumer_group_instance_id,
+                group_instance_id=conf.consumer_group_instance_id,
                 bootstrap_servers=server_list(transport.url, transport.default_port),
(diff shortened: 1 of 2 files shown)


wbarnha (approved):
No issue on my end, now that it's supported. I'll defer to the opinion of Vikram since he knows more about static assignment, in case there are any caveats.

tests/unit/transport/drivers/test_aiokafka.py
```diff
@@ -787,7 +787,7 @@ def assert_create_worker_consumer(
                 api_version=app.conf.consumer_api_version,
                 client_id=conf.broker_client_id,
                 group_id=conf.id,
-                # group_instance_id=conf.consumer_group_instance_id,
+                group_instance_id=conf.consumer_group_instance_id,
```
patkivikram, line 790: is this passed in from the application? As long this is passed in we are good here

wbarnha: https://github.com/faust-streaming/faust/blob/e91c55fea51f77306da2b2a0e1e3718c34081aa3/faust/types/settings/settings.py#L1106-L1119

Looks like it's been passed in for some time, just never got used.

PR #608, opened 2024-02-15
bugfix: Fix cython _on_stream_event_in call to match python impl

### LINKED ISSUE #607: Logic around sensor _on_stream_event_in differs between cython and python implementations
## Checklist

- [X] I have included information about relevant versions
- [X] I have verified that the issue persists when using the `master` branch of Faust.

I recently added prometheus metrics instrumentation to one of my teams projects that's using faust-streaming. However, many of the per-agent metrics I was expecting to appear did not appear. After extensive debugging, I realized that this is due to agents using `yield` and `sink` were not correctly invoking _on_stream_event_in per event.

I dug through the cython and python implementations of the stream handling and discovered that in cython, `_on_stream_event_in` is being invoked conditionally alongside `_on_message_in`:
https://github.com/faust-streaming/faust/blob/master/faust/_cython/streams.pyx#L177-L183
but in the python implementation, the `_on_stream_event_in` is being invoked all the time in this code path:
https://github.com/faust-streaming/faust/blob/master/faust/streams.py#L1182-L1190

These cannot both be correct. I believe the correct behavior is the python one.

## Expected behavior

I expect that _on_stream_event_in should be called per stream event in both the cython and python impls.

## Actual behavior

_on_stream_event_in is not invoked per stream event when using the cython implementation.

Tim-Thomas:
## Description
Changes the cython logic around invoking `_on_stream_event_in` to match the logic in the python implementation.

Fixes #607


wbarnha:
Thanks for the PR! Just run `./scripts/lint` and you should be good to go for linting.

Tim-Thomas:
> Thanks for the PR! Just run `./scripts/lint` and you should be good to go for linting.

Running that script changes a large number of files so I opened another PR to match. If this is an issue with my local black config or flake8, please let me know and I'll take another look. I am assuming that there was a change in black as the linting failures that appear in the CI/CD job match the changes that I saw locally when running linting.

PR here: https://github.com/faust-streaming/faust/pull/609

wbarnha (approved):
https://github.com/faust-streaming/faust/blob/6d248bf5e0ab7ac5961bad76839149ac03f7bb86/faust/streams.py#L1183-L1191

Indeed it should be, good catch!

PR #609, opened 2024-02-16
Run black and update setup.cfg for flake8 to ignore E701 and E704

Tim-Thomas:
This re-formats everything to match current black configuration and allow flake8 to succeed.


Tim-Thomas:
This seems to be a deliberate change to black: https://github.com/psf/black/pull/4066/files and the conflict with flake8 as well as the solution is documented here: https://github.com/psf/black/issues/4173

wbarnha (approved):
Thank you for this PR as well! LGTM!

Issue #175, opened 2021-07-21
Consumers slowly die over time


pb376:
Running the application for any extended period of time - sometimes just a few minutes - leads to consumers slowly dying one-by-one until they've all stopped responding and the entire application hangs. The data is still flowing into a topic at the tip of the pipeline from an external source, so there are still messages to be read as they begin to hang.

The following messages are displayed in increasing numbers as things begin to die:

```
[2021-07-21 01:04:39,262] [31912] [ERROR] [^---AIOKafkaConsumerThread]: Stream has not started processing TP(topic='archive-log-events', partition=36) (started 1.10 hour ago). 
                                                                                                                                          
There are multiple possible explanations for this:                                                                                                
                                                                                                                                                                                              
1) The processing of a single event in the stream                                                                                    
   is taking too long.                                                                                                                    
                                                                                                                                          
    The timeout for this is defined by the stream_processing_timeout setting,                                                             
    currently set to 300.0.  If you expect the time                                                                                                                                               required to process an event, to be greater than this then please                                                                                            
    increase the timeout.                                                                                                                                                                     
                                                                                                                                          
 2) The stream has stopped processing events for some reason.                                                                       
                                                                                                                                          
3) The agent processing the stream is hanging (waiting for network, I/O or infinite loop).  
```

I've tested this with agents performing simple functions and the behavior still shows. Is there any obvious reason why this would be occurring or a reasonable way to debug it? 

## Checklist

- [X] I have included information about relevant versions
- [X] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

Run an application with multiple agents.

## Expected behavior

Things run indefinitely.

## Actual behavior

Things run and slowly begin stop.

## Full traceback

N/A

# Versions

* Python version: `3.8.6`
* Faust version:  `0.6.9`
* Operating system: `Ubuntu 18.04`
* Kafka version: `2.8.0`
* RocksDB version (if applicable)

mavalderrama:
we have the same issue right now, there is no clear error detail on whats going on

PJ-Schulz:
Have you tried to increase the broker_commit_livelock_soft_timeout or the broker_max_poll_interval?

mavalderrama:
> Have you tried to increase the broker_commit_livelock_soft_timeout or the broker_max_poll_interval?

thanks, I'll try it in production

oneryalcin:
yes I noticed this behaviour too. Anybody has a solution to this?

[...]

PR #610, opened 2024-02-23
Fix disabled aiokafka unit tests

wbarnha:
It's going to be hard to investigate what's going on with https://github.com/faust-streaming/faust/issues/175 unless we get better testing in place.

### DIFF
--- a/faust/transport/drivers/aiokafka.py
+++ b/faust/transport/drivers/aiokafka.py
@@ -840,14 +840,14 @@ class AIOKafkaConsumerThread(ConsumerThread):
         poll_at = None
         aiotp_state = assignment.state_value(aiotp)
         if aiotp_state and aiotp_state.timestamp:
-            poll_at = aiotp_state.timestamp / 1000
+            poll_at = aiotp_state.timestamp
         if poll_at is None:
             if secs_since_started >= self.tp_fetch_request_timeout_secs:
                 # NO FETCH REQUEST SENT AT ALL SINCE WORKER START
                 self.log.error(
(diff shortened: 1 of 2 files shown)


faust/transport/drivers/aiokafka.py
```diff
@@ -840,14 +840,14 @@ def _verify_aiokafka_event_path(self, now: float, tp: TP) -> bool:
         poll_at = None
         aiotp_state = assignment.state_value(aiotp)
         if aiotp_state and aiotp_state.timestamp:
-            poll_at = aiotp_state.timestamp / 1000
```
wbarnha, line 843: Looks like this change was originally implemented in https://github.com/faust-streaming/faust/commit/dcc421fc29a8b555ff38a0bfac4366fcc04d2369 by @patkivikram, could you expand on this?

faust/transport/drivers/aiokafka.py
```diff
@@ -840,14 +840,14 @@ def _verify_aiokafka_event_path(self, now: float, tp: TP) -> bool:
...
+            poll_at = aiotp_state.timestamp
         if poll_at is None:
             if secs_since_started >= self.tp_fetch_request_timeout_secs:
                 # NO FETCH REQUEST SENT AT ALL SINCE WORKER START
                 self.log.error(
                     SLOW_PROCESSING_NO_FETCH_SINCE_START,
                     tp,
-                    secs_since_started,
+                    humanize_seconds_ago(secs_since_started),
```
wbarnha, line 850: Accidentally reverted this in `master`, ignore.

PR #613, opened 2024-03-04
Fix prometheus registry

krzysieksulejczak:

## Description

The `setup_prometheus_sensors` function accepts `registry` parameter with default set to default prometheus registry. The web view to render metrics explicitly uses prometheus default registry. If anyone uses non-default registry in call to `setup_prometheus_sensors` the metrics will not be available through web view.


wbarnha (approved):
Good catch, LGTM!

Issue #630, opened 2024-07-02
Error when creating a topic using aiokafka 0.11.0


hgalytoby:
poetry add faust-stream

```py
import faust


class Greeting(faust.Record):
    from_name: str
    to_name: str


app = faust.App('hello-app', broker='kafka://localhost')
topic = app.topic('hello-topic', value_type=Greeting)


@app.agent(topic)
async def hello(greetings):
    async for greeting in greetings:
        print(f'Hello from {greeting.from_name} to {greeting.to_name}')


@app.timer(interval=1.0)
async def example_sender(app):
    await hello.send(
        value=Greeting(from_name='Faust', to_name='you'),
    )


if __name__ == '__main__':
    app.main()
```

### error
`AttributeError: 'CreateTopicsRequest_v1' object has no attribute 'build_request_header' `

No problems with aiokafka 0.10.0

fonty422:
Do you need the Faust application to create the topic? Can you try to add the topic manually with the Kafka topic scripts and see if that helps?

hgalytoby:
Usually I don't need faust to create a topic, usually faust just creates the leader topic for me.

I know there are other ways to create topics, it's just a matter of me finding out about it and letting it be known!

fonty422:
I have an issue with my applications where if I use Faust to create the topics it often results in a topic not being created (something about topic can't be used while being created) and then the error message ignoring missing topic: <topic name>. I think this is due to a large volume of applications being started at the same time and the CPU is wildly overloaded.

To get around this, I manually create the topics before hand using the kafka_topics.sh (or .bat if using windows) that's part of your kafka download and found in the bin directory.

For me, when ever I start my application I first run:
`kafka-topics.sh --create --if-not-exists --topic <topic name> --bootstrap-server=<address> --partitions=<partitions>`
replacing topic name, address, and partitions with the appropriate values.

The --if-not-exists argument will only create the topic if it doesn't already exist.

jjwwiodyne:
I'm seeing the same issue with the combination of faust 0.11.0 and aiokafka 0.10.0. Forcing aiokafka==0.10.0 fixes the issue.

somnam:
Added a PR for this https://github.com/faust-streaming/faust/pull/631, can someone have a look?

PR #631, opened 2024-07-09
fix: aiokafka `0.11.0` error when creating topics

somnam:

## Description
Please note this build _wasn't_ verified on an example Faust app.

Aiokafka introduced `build_request_header` and `parse_response_header` calls in `conn.py` module in `0.11.0` release:
https://github.com/aio-libs/aiokafka/commit/a244a5f575d06cf7d0f8e4ae21847a4d768bb0e7#diff-60fe72976f73ef2b8b881a26cc5d9e002c465e7b1b7f1b83e0be07c672a66e0e

Use `CreateTopicsRequest`  from `aiokafka.protocol.admin` module instead of `faust.utils.kafka`.
Remove `faust.utils.kafka` module containing what looks like older version of code from `aiokafka.protocol.admin`.

Fixes https://github.com/faust-streaming/faust/issues/630

> AttributeError: 'CreateTopicsRequest_v1' object has no attribute 'build_request_header' 

that version series.


wbarnha (approved):
Thanks for the PR! I've been wanting to move a lot of Kafka internals to be more dependent on `aiokafka` since the project is in a more developed state than when Faust was originally made. 

Also, thank you for updating the test to accommodate your change! Looks all good here, and I'll approve it. https://github.com/aio-libs/aiokafka/blob/199537bb02177ff935ccdefa2b44615214024a0f/aiokafka/protocol/admin.py#L201 is analogous to the original implementation in Faust, so this is a safe merge.

wbarnha:
The `distutils` issue will be handled in a separate PR.

PR #635, opened 2024-07-16
Replace aredis with redis-py

wbarnha:
We need to get `aredis` out of Faust since it uses `distutils`, a library removed in Python 3.12.


Luferov:
Faust doesn't work with ardis, because aredis support by python until 3.8. Miracle, python 3.9 is working with aredis, Faust with 3.11 doesn't work. We move with aredis to redis-py 🙏

wbarnha:
I'm waiting to merge this until I've tested this actual an actual Redis instance.

wbarnha:
We're going to have to get this merged if we want to support newer versions of Python.

Issue #634, opened 2024-07-16
Worker logging doesn't work


lpogo:
After upgrade to faust-streaming > 0.8.10 (PR: https://github.com/faust-streaming/faust/pull/368) I am not able to force faust to print info like: 
```
[^Worker]: Starting...
[^-App]: Starting...
[^-App]: Started.
Executing _on_partitions_assigned
[^Worker]: Ready
[^Worker]: Started.
```
..as well as each processed event. Before it was working like that. 
Now I just receive `+ƒaµS† v0.8.11+` table info and nothing after that. The same for the newest version. 
Any idea whats wrong or how to handle logging?
When I am using logging_config in faust.App my logging configuration in version 0.8.10 works fine. Since 0.8.11 it is not.

I am using faust-streaming with Django 4.2.11

PR #368, opened 2022-09-22
Fix autodiscovery error log not showing up

blazing-gig:
*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://faust-streaming.github.io/faust/contributing.html).

## Description

Fixes #293

PR #636, opened 2024-07-17
Add override_logging argument to Worker (#634)

lpogo:
* Add override_logging argument to Worker 
* Add worker_override_logging to App
* Update documentation
* Update tests

## Description

Allow user to decide if override_logging shoud be enabled or not.
It is forced by faust.Worker to be enabled right now.

It resolve Issue #634

### DIFF
--- a/docs/includes/settingref.txt
+++ b/docs/includes/settingref.txt
@@ -2135,6 +2135,21 @@ Level used when redirecting standard outputs.
 The logging level to use when redirect STDOUT/STDERR to logging.
 
 
+.. setting:: worker_override_logging
+
+``worker_override_logging``
+---------------------------
+
+:type: :class:`bool`
+:default: :const:`True`
+:environment: :envvar:`WORKER_OVERRIDE_LOGGING`
+
+Override worker logging.
+
+Enable to configure root logger.
+
+Enabled by default.
+
 
 .. _settings-extending:
 
--- a/faust/cli/base.py
+++ b/faust/cli/base.py
@@ -516,6 +516,7 @@ class Command(abc.ABC):  # noqa: B024
     daemon: bool = False
     redirect_stdouts: Optional[bool] = None
     redirect_stdouts_level: Optional[int] = None
+    override_logging: Optional[bool] = None
 
     builtin_options: OptionSequence = builtin_options
     options: Optional[OptionList] = None
@@ -660,6 +661,7 @@ class Command(abc.ABC):  # noqa: B024
             console_port=self.console_port,
             redirect_stdouts=self.redirect_stdouts or False,
             redirect_stdouts_level=self.redirect_stdouts_level,
+            override_logging=self.override_logging or False,
             loop=loop or asyncio.get_event_loop_policy().get_event_loop(),
             daemon=self.daemon,
         )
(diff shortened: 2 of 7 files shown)


lpogo:
@patkivikram, @wbarnha could you please approve it to run workflow? Thanks!

wbarnha (approved):
Thanks for updating the tests as well! LGTM!

davidvanwyk:
@wbarnha - this has seemingly broken logging for me. Was working fine on 0.11.1 and now after this change I get no logging :/

wbarnha:
I feared this would be the case. Looks like I missed this during testing. Does setting `worker_override_logging=False` in your settings fix your issue? I was contemplating setting that as the default.

davidvanwyk:
@wbarnha - I haven't set up any form of custom logging. I'm using the defaults with no overrides and I invoke the worker as follows: 

```
python -m my_app worker -l info
```

I get the exact reverse behavior of the OP (what they were seeing in their linked issue I am now seeing after this change)

wbarnha:
Well, that's a shame... I'll try to find a happy medium.

lpogo:
@davidvanwyk could you confirm, that if you set `override_logging=True` in your `my_app` application, then logging is working fine again?
If yes, I think we should use `True` instead of `False` as default value. `mode.Worker` is using `True` as default, so it may be my bad I set it to `False`.
@wbarnha what do you think?

davidvanwyk:
@lpogo - unfortunately not. I still get no logs other than the initial stdout with the Faust version etc.

davidvanwyk:
Just to confirm - the setting you enabled on the application instance is: `worker_override_logging`. I think `override_logging` was what it was named in the context of CLI option?

lpogo:
> Just to confirm - the setting you enabled on the application instance is: `worker_override_logging`. I think `override_logging` was what it was named in the context of CLI option?

Yes, you are right.

davidvanwyk:
Well take a look and see if you can resolve this in the near term and @wbarnha can just roll forward, otherwise I think we need to roll this one back until it's been figured out? :) If I have some time in the coming days I can check it out as well but not 100% sure I will have the time.

wbarnha:
I agree, I'm a bit confused on why this feature was necessary to begin with but I merged anyway since I figured it'd be harmless. We'll roll it back.

lpogo:
Lets do it this way. This change mustn't impact current users. 
@wbarnha I found that it doesn't matter if `worker_override_logging` is set to `True` - `faust.Worker` is always receiving `False`. No idea why it is happening now..
Anyway for me logging issue started after #368 . Please take a look there.

lesmanacflt:
This one hit me as well. Seems like now logging doesn't work. Rolling back to previous version.

wbarnha:
Reverted changes in `faust-streaming==0.11.3`, thanks for your patience!

wbarnha:
> Lets do it this way. This change mustn't impact current users. @wbarnha I found that it doesn't matter if `worker_override_logging` is set to `True` - `faust.Worker` is always receiving `False`. No idea why it is happening now.. Anyway for me logging issue started after #368 . Please take a look there.

I acknowledge the issue you're facing, we'll need to solve this some other way.

PR #658, opened 2025-03-17
Also use faust rebalance protocol when using static parition assignment

Peter200lx:
In 2022 Faust added a fallback to Round Robin w/ PR #402 when `table_standby_replicas` wasn't used. In 2024 Faust added support for KIP-345 kafka static-partition-assignment #600. When a user is specifying static partition assignment, using Round Robin doesn't respect the static partition instance IDs. This re-adds use of the Faust assigner when static partition assignment is configured, and brings it in line with comment on `consumer_group_instance_id`:
https://github.com/faust-streaming/faust/blob/master/faust/types/settings/settings.py#L1113-L1116
```
        The group_instance_id for static partition assignment.

        If not set, default assignment strategy is used. Otherwise,
        each consumer instance has to have a unique id.
```
Also the variable defaults to `None`, thus the truthiness check being a simple way of testing if it has been set:
https://github.com/faust-streaming/faust/blob/master/faust/types/settings/settings.py#L113
```
        consumer_group_instance_id: Optional[str] = None,
```


wbarnha (commented):
This is definitely a good idea, but it would give me peace of mind knowing that there is a test checking that this works.

PR #660, opened 2025-05-16
fixed StreamT annotations, TPorTopicSet can also be optional

tworedz:

## Description

* Fixed / added type annotations mostly for `StreamT` class. 
* Fixed generic argument for `EventT` class.
* Make sure that App.commit accepts `None` as a documentation saying (https://faust-streaming.github.io/faust/reference/faust.app.html#faust.app.App.commit)
* Removed `__iter__/__aiter__` methods on `StreamT`. And it makes sense, because it is extending `AsyncIterable`, which already implements `__aiter__`. `__iter__` was just redundant here and broke typing.


tworedz:
please review @wbarnha

wbarnha (commented):
Thanks for the PR! Most of the changes look good for merge, I am a bit surprised that `TPorTopicSet` was not already annotated with `Optional`. I will look closer at the code to make sure this is proper.

tworedz:
Thanks! Seems that pypy doesn't like my changes, btw. Is it because of me or they were failing before?

wbarnha:
I'll come back to this soon. Thank you for your work.

PR #673, opened 2026-02-25
fix: remove outdated stop_and_shutdown_loop from Worker docstring

### LINKED ISSUE #317: Worker docstring incorrect, `stop_and_shutdown_loop` doesn't exist anymore
I want to run the `Worker` class from my code. There's a [docstring on the class](https://github.com/faust-streaming/faust/blob/master/faust/worker.py#L172) explaining how to do it. It looks to be outdated, though, as there's no `stop_and_shutdown_loop` method on the worker anymore.

If I'm reading the code correctly, the Worker will register signal handlers on start, so there's no need for any cleanup in `finally`, right?

P.S. I'm running `Worker` from my own code, so that I can retry connecting to Kafka a couple of times. While I'm starting up my containers with Docker Compose, Kafka might need some time to set up, and my app needs to wait for it.

nightcityblade:
## Description

Fixes #317

The Worker class docstring referenced `stop_and_shutdown_loop()` in example 3, but this method no longer exists. The Worker registers signal handlers on start, so explicit cleanup in a `finally` block is unnecessary.

Removed the outdated `try/finally` block and the reference to `stop_and_shutdown_loop()` from the docstring example.

### DIFF
--- a/faust/worker.py
+++ b/faust/worker.py
@@ -156,9 +156,7 @@ class Worker(mode.Worker):
                     >>> worker = Worker(app)
                     >>> worker.execute_from_commandline()
 
-            3) or if you already have an event loop, calling ``await start``,
-               but in that case *you are responsible for gracefully shutting
-               down the event loop*::
+            3) or if you already have an event loop, calling ``await start``::
 
                     async def start_worker(worker: Worker) -> None:
                         await worker.start()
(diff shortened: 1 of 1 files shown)


nightcityblade:
Gentle ping — any thoughts on this? Happy to address feedback if needed.

PR #676, opened 2026-03-31
fix(aiokafka): keep Faust assignor for changelog tables and update admin request compatibility

Mopsgeschwindigkeit:
## Description

Ensure worker consumers use the Faust assignor whenever changelog tables are present, even with zero standby replicas, so table recovery mapping remains correct. Also update metadata/create-topics request construction for newer aiokafka APIs.

Also removed support for python3.9 and below.

## Problems

- If `self.app.conf.table_standby_replicas = 0` the `RoundRobinPartitionAssignor` is used even is the app contains `changelog_topics`, which leads to a situation where `changelog_topics` are not consumed at the start of the app.

- With newer version of aiokafka, some older api-version where deprecated.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -42,13 +42,9 @@ jobs:
       # for example if a test fails only when Cython is enabled
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
         use-cython: ['true', 'false']
         experimental: [false]
-        include:
-          - python-version: 'pypy3.9'
-            use-cython: false
-            experimental: true
     env:
       USE_CYTHON: ${{ matrix.use-cython }}
     continue-on-error: ${{ matrix.experimental }}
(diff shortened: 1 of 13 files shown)


Mopsgeschwindigkeit:
@dada-engineer @patkivikram that one test "pypy3.9" fails during installation. Since python3.9 is EOL since end of last year. Could we remove support for python3.9 ?

dada-engineer:
> @dada-engineer @patkivikram that one test "pypy3.9" fails during installation. Since python3.9 is EOL since end of last year. Could we remove support for python3.9 ?

I think so yes. Ideally we make sure it is python 3.14 tested then

Mopsgeschwindigkeit:
> > @dada-engineer @patkivikram that one test "pypy3.9" fails during installation. Since python3.9 is EOL since end of last year. Could we remove support for python3.9 ?
> 
> I think so yes. Ideally we make sure it is python 3.14 tested then

Just had a look, python3.14 does not work out-of-the-box. So a separate PR would be a better option then adding it here.

dada-engineer:
> Just had a look, python3.14 does not work out-of-the-box. So a separate PR would be a better option then adding it here.

Makes sense.

Mopsgeschwindigkeit:
seems a like a maintainer has to remove python3.9 from the required checks.

Mopsgeschwindigkeit:
@dada-engineer any news ?

PR #689, opened 2026-07-18
Native OpenTelemetry tracing — foundation + ADR (Phase 1a)

wbarnha:
## Description

Groundwork for migrating Faust's built-in distributed tracing off the **archived / EOL OpenTracing API** to **native OpenTelemetry**, while keeping the full processing-pipeline span tree (agents, streams, table recovery, assignor, timers, Crontabs, rebalance).

This is **Phase 1a — foundation only**. It adds an architecture decision record and a self-contained, unit-tested native tracing core. **No existing call sites are rewired yet**, so the running OpenTracing layer is completely unaffected. Opening as a **draft** to get agreement on the approach before the larger Phase 1b rewire.

> **Synced with master.** Master has since landed the sensor-level OpenTelemetry path (#748), the OTel metrics monitor (#746), and moved `opentracing` out of core requirements into the `faust[opentracing]` extra (#686). The merge commit resolves the `requirements/extras/opentelemetry.txt` and `setup.py` conflicts and refreshes the ADR against the new tree — see "Relationship to `faust.contrib.opentelemetry`" in the ADR.

### Why native OTel (and why not an instrumentor)

- OpenTracing is archived/EOL; the OpenTelemetry OpenTracing shim was formally deprecated on 2026-03-19 (removal ≥ Mar 2027). The shim bridge (#688) is a valid interim, not a destination.
- There is **no `opentelemetry-instrumentation-faust`** package — not on PyPI, not in `opentelemetry-python-contrib`, nothing community-maintained. The transport-level instrumentors (`opentelemetry-instrumentation-aiokafka` / `-confluent-kafka`, both beta) only patch the raw Kafka clients and cannot reproduce Faust's agent/stream/recovery/**rebalance** spans — they complement, they can't replace.
- A library should depend only on `opentelemetry-api` (no SDK); calls are cheap no-ops until the app registers a `TracerProvider`. `opentelemetry.context` is contextvars-based, so Faust's current-span propagation maps 1:1, including across `await`.

### Relationship to `faust.contrib.opentelemetry` (#748)

Complementary, not competing. `OpenTelemetrySensor` sits at the **sensor** layer — one `{topic} process` span per event, which is what closes the aiokafka instrumentor's cross-thread gap — and it gives users native traces today without touching `app.tracer`. Sensors have no hooks for table recovery, the assignor, timers, Crontabs, the rebalance sequence, or the produce side; this work covers that **built-in span tree**, the ~134 call sites that today can only emit OpenTracing spans. After Phase 1b both emit into the same native OTel trace.

### What's in this PR

- **`docs/adr/0001-native-opentelemetry-tracing.md`** — the decision record: options considered, how this relates to the sensor-level path, semantic-convention coexistence, native re-expression of the two non-portable OpenTracing hacks, and the phased migration plan.
- **`faust/utils/otel_tracing.py`** — api-only tracing primitives that degrade to no-ops when OpenTelemetry is absent:
  - tracer resolution against the global provider;
  - contextvar current-span helpers;
  - **`deterministic_parent_context()`** — native replacement for the old `span.context.trace_id = murmur2(...)` mutation (OTel span contexts are immutable): a `NonRecordingSpan` parent carries the seeded trace id and the real rebalance span inherits it;
  - **`PendingSpan`** — native replacement for the `_transform_span_lazy` `span.finish` monkeypatch: defer span creation until the Kafka generation is known, or cancel;
  - **semconv-coexistence helpers** — every span carries both the legacy `kafka-*` tags (deprecated) and the OpenTelemetry `messaging.*` attributes + `SpanKind`;
  - W3C `traceparent` header inject/extract.
- **`tests/unit/utils/test_otel_tracing.py`** — 19 tests covering every primitive against an in-memory SDK exporter (both hacks validated end-to-end); the no-OTel degradation path is also exercised.
- **`requirements/extras/opentelemetry.txt`** — adds `opentelemetry-sdk` to the existing extra, so apps can configure an exporter out of the box. The tracing core itself needs only `opentelemetry-api`.

### Semantic-convention coexistence

Adopted **additively**: span *names* stay legacy by default (`consume-from-{topic}`, `job-{topic}`, `produce-to-{topic}`) and are marked deprecated, while `messaging.*` attributes + `SpanKind` coexist on the same span. A future opt-in flips names to semconv (`{topic} process`, …).

### Migration plan (see ADR)

- **Phase 0 (done):** `opentracing` made optional (#686), shim bridge (#688), OTel metrics monitor (#746), sensor-level tracing (#748) — all non-breaking.
- **Phase 1a (this PR):** native tracing core + ADR, nothing rewired.
- **Phase 1b:** convert the 12 modules; redefine `TracerT` in OTel terms; ship a legacy adapter wrapping existing OpenTracing `app.tracer` implementations via the shim for one major cycle. **Targets a major version** — `app.tracer`/`TracerT` is a breaking signature change.
- **Phase 2:** remove the adapter and retire the `faust[opentracing]` extra along with `faust/utils/_opentracing.py`.

### Notes for reviewers

- Nothing here changes runtime behavior — the new module isn't imported by Faust core yet.
- The breaking-change and release-timing decisions (Phase 1b) are intentionally left for maintainers before that work lands.


wbarnha:
Closing as a duplicate of #782, which is a master-synced resubmission of this same branch (identical files: docs/adr/0001-native-opentelemetry-tracing.md, faust/utils/otel_tracing.py, requirements/extras/opentelemetry.txt, tests/unit/utils/test_otel_tracing.py).

PR #719, opened 2026-07-20
test: re-enable recoverable skipped tests (+ fix latent LazySpan bug)

wbarnha:
## Description

Follow-up to #716. Sweeps the suite's 20 legacy unconditional `@pytest.mark.skip("Needs fixing"/"Fix is TBD")` tests, re-enabling every one that can be made green and **fixing a real bug** a skip was hiding. Stacked on `claude/pypy-agen-cleanup` (#716) — retarget to `master` once that merges.

### Re-enabled (7 tests)

- **Real bug fix — `faust/transport/drivers/aiokafka.py`:** `_transform_span_lazy` assigned the raw, **unbound** `LazySpan.finish` to `span.finish`, so the later `span.finish()` raised `TypeError: finish() missing 1 required positional argument: 'self'`. Lazy spans are used whenever tracing is enabled, so this was a latent crash — the four skipped `test_transform_span_*` tests were masking it. Bound the method to the span; **4 tests re-enabled.**
- **`test_aiokafka.py` stream-idle VEP (2):** `_make_slow_processing_error` grew `setting`/`current_value` params and now bakes the explanation into the message, so `log.error` no longer gets those as kwargs. Updated the two stale expectations.
- **`test_agent.py::test_execute_actor__cancelled_running` (1):** "Fix is TBD" — passes as-is now (skip rot).

### Attempted, still skipped — but with precise reasons (was "Needs fixing")

- **`test_aiokafka.py`** fetch-timeout VEP asserts ×3 — `verify_event_path` no longer emits the error() under those conditions (driver behaviour drift); `_commit` mock tests ×3 — offsets no longer flow to the wrapped `AIOKafkaConsumer.commit` in that harness. Both need real test re-derivation.

### Attempted, left as-is (need infra / dependency-API work)

Documented here rather than re-skipped in code:

| Test(s) | Why it can't just be re-enabled |
|---|---|
| `test_cache.py` redis scheme (2) | Attempt a live Redis connection; redis-py removed `StrictRedis` |
| `test_rocksdb.py::test_open_rocksdict` | `rocksdict`'s `Options()` no longer takes constructor args |
| `test_consumer.py` TransactionManager (2) | `stop_transaction` call-count / expected-calls drift |
| `test_base.py::test_compat_option` | CLI compat-option callback path changed |
| `test_consistency.py` | Module import fails on current aiokafka (`GroupCoordinatorRequest_v0` removed); this dir isn't in the CI suite |

### Verification

`tests/unit/transport` + `tests/unit/agents/test_agent.py` — **598 passed, 16 skipped, 0 failed** on CPython; flake8/black/isort clean.

---

### DIFF
--- a/faust/transport/drivers/aiokafka.py
+++ b/faust/transport/drivers/aiokafka.py
@@ -629,7 +629,14 @@ class AIOKafkaConsumerThread(ConsumerThread):
                 def finish(self) -> None:
                     consumer._span_finish(span)
 
-            span._real_finish, span.finish = span.finish, LazySpan.finish
+            # Bind the replacement to ``span`` -- assigning the raw
+            # ``LazySpan.finish`` function leaves it unbound, so the later
+            # ``span.finish()`` call raises "missing 1 required positional
+            # argument: 'self'".
+            span._real_finish, span.finish = (
+                span.finish,
+                LazySpan.finish.__get__(span),
+            )
 
     def _span_finish(self, span: opentracing.Span) -> None:
         assert self._consumer is not None
--- a/tests/unit/agents/test_agent.py
+++ b/tests/unit/agents/test_agent.py
@@ -469,7 +469,6 @@ class Test_Agent:
             await agent._execute_actor(coro, Mock(name="aref", autospec=Actor))
         coro.assert_awaited()
 
-    @pytest.mark.skip(reason="Fix is TBD")
     @pytest.mark.asyncio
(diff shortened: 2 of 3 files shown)


wbarnha:
Closing in favour of #723, which now contains this work rebased directly onto `master` (independent of the PyPy branch it was previously stacked on) plus the follow-up aiokafka/CLI re-enablements. No content is lost — it all lives in #723.

---

PR #721, opened 2026-07-20
test(redis cache): re-enable Test_RedisScheme + not-installed test for redis-py 8.x

wbarnha:
## Description

Re-enables three redis cache tests skipped *"Needs fixing"* that had drifted against redis-py 8.x and the current cache backend.

- **`Test_RedisScheme::test_single_client` / `test_cluster_client`** referenced `redis.StrictRedis` / `aredis.RedisCluster` as attributes of the backend module — which no longer exist there. They now assert against the scheme→client-class map the backend actually builds (`_client_by_scheme`): single-node resolves to redis-py 8.x's concrete `redis.client.Redis`; cluster resolves to `redis.RedisCluster`. The single-node client is built **for real** (redis-py connects lazily) and its `connection_pool.connection_kwargs` (host/port/db) are checked; the cluster class — which connects on construction — is mocked exactly like the existing `mocked_redis` fixture, while scheme selection and `db`-stripping run for real. `test_single_client` is strictly stronger than the original.
- **`test_redis__aredis_is_not_installed`** now patches the `redis` module name the backend's `on_start` guard actually reads (`if redis is None`), so it raises `ImproperlyConfigured` without opening a socket (the old fixture patched `aredis`, which the guard never checks — so it opened a real connection to `localhost:6079`).

Only external redis client classes are mocked; the backend logic runs for real. **Test-only change.**

### Latent source bug noticed (not fixed here)
The backend does `import redis.asyncio as aredis` but its `on_start` guard only checks `if redis is None`, never `aredis`. So the "async redis not installed" path is effectively dead / guards the wrong name. Left for a separate source-side PR to keep this change test-only.

## Verification
`tests/functional/web/test_cache.py` → **32 passed**; flake8/black/isort clean. Produced under adversarial faithfulness review.

---

### DIFF
--- a/tests/functional/web/test_cache.py
+++ b/tests/functional/web/test_cache.py
@@ -1,7 +1,6 @@
 from itertools import count
 
 import pytest
-import redis.asyncio as aredis
 
 import faust
 from faust.exceptions import ImproperlyConfigured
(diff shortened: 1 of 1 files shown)


wbarnha:
Closing in favour of #724, which now contains this test rewrite **plus** the backend not-installed-guard source fix, as a single independent PR off `master`. No content is lost — it all lives in #724.

---

PR #724, opened 2026-07-20
redis cache: re-enable scheme/not-installed tests + fix not-installed guard

wbarnha:
## Description

Re-enables the skipped redis cache tests **and** fixes the latent backend bug they surfaced. **Independent, off `master`** (consolidates the former #721 + the source fix; #721 is closed in favour of this).

### Source fix — `faust/web/cache/backends/redis.py`

The import guard bound only `aredis` on failure:

```python
except ImportError:
    aredis = None        # <-- redis left unbound
```

but the backend guards on `if redis is None` (`on_start`, scheme/error setup). With redis genuinely uninstalled, `redis` stays *unbound* and the first `if redis is None` raises `NameError` at import instead of the intended `ImproperlyConfigured`. Fixed by binding **`redis = None`**, and dropped the **dead `import redis.asyncio as aredis`** alias (only ever assigned `None`, never referenced — its presence is what made the guard look correct).

### Tests — `tests/functional/web/test_cache.py`

- `Test_RedisScheme::test_single_client` / `test_cluster_client` — rewritten for redis-py 8.x: assert against the scheme→client map the backend actually builds (`_client_by_scheme`); single-node resolves to `redis.client.Redis` (built for real; connects lazily), cluster to `redis.RedisCluster` (mocked like the existing `mocked_redis` fixture, since it connects on construction). Scheme selection and `db`-stripping run for real.
- `test_redis__is_not_installed` (renamed from `…aredis_is_not_installed`) — patches the `redis` name the guard reads and asserts `ImproperlyConfigured` without opening a socket.

### Verification
`tests/functional/web/test_cache.py` → **32 passed**; the guard now raises `ImproperlyConfigured` (previously `NameError`), verified directly. flake8/black/isort clean.

### Not addressed (separate, larger concern)
The backend maps schemes to **sync** clients (`redis.StrictRedis`/`RedisCluster`) yet `await`s them in `_get`/`_set` — fine under the mocked tests, likely broken for real async use. Worth a dedicated look.

### DIFF
--- a/faust/web/cache/backends/redis.py
+++ b/faust/web/cache/backends/redis.py
@@ -16,12 +16,15 @@ from . import base
 
 try:
     import redis
-    import redis.asyncio as aredis
     import redis.exceptions
 
     redis.client.Redis
 except ImportError:  # pragma: no cover
-    aredis = None  # noqa
+    # ``on_start`` (and the module-level ``if redis is None`` guards) key off
+    # ``redis`` being ``None`` when the library is missing.  The name must be
+    # bound to ``None`` here -- otherwise it stays unbound and the guard below
+    # raises ``NameError`` instead of the intended ``ImproperlyConfigured``.
+    redis = None  # noqa
 
 if typing.TYPE_CHECKING:  # pragma: no cover
     from redis import StrictRedis as _RedisClientT
--- a/tests/functional/web/test_cache.py
+++ b/tests/functional/web/test_cache.py
@@ -1,7 +1,6 @@
 from itertools import count
 
 import pytest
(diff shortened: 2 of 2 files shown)


wbarnha:
Superseded by #725. Investigating "can we test redis in CI" revealed the backend is actually broken against a real server (it `await`s the sync `redis.StrictRedis` → `TypeError`), which this PR's mocks were hiding. #725 folds in this guard fix, switches the backend to the `redis.asyncio` clients so it works for real, updates the scheme tests accordingly, and adds a live-redis CI job. Closing in favour of it — nothing is lost.

---

PR #748, opened 2026-08-04
Pluggable web frameworks, OpenTelemetry tracing, and web feature flags

### LINKED ISSUE #322: Faust hangs on publish_message unpredictabilly
## Steps to reproduce

We are using faust-streming within our fastAPI application and start faust on fastAPI event startup with:

    asyncio.create_task(faust_app.start())

Unfortunatelly after some time (as of now non deterministic so I can not provide a reproducable example sorry) we are not able to consume or produce messages. However in the producer we do see the debug log publish_message called with key=b'xxx' and value=b'{"some":"json"}', but the message is not sent to Kafka. Nevertheless we do see mode.timers 

We run only one broker, one faust application, on a topic with one partition. And suspect some rebalincing issues.

Are there any insights on this yet?

## Expected behavior

Faust should not hang.

## Actual behavior

Faust hangs without any error

## Full traceback

There is none.

# Versions

* Python version: 3.9.x
* Faust version: 0.8.4 and 0.8.5 (maybe also earlier, maybe config issue on our side)
* Operating system: Debian
* Kafka version: 3.1.0
* RocksDB version (if applicable): 6.29.3 (shouldn't be an issue here though)

Any help and hints are highly appreciated, thank you.

### LINKED ISSUE #435: ValueError: The future belongs to a different loop than the one specified as the loop argument

### LINKED ISSUE #448: Can't use FastAPI + Faust-streaming version above 0.10.1 - attached to a different loop
Thanks for all the people continuing to support this project, unfortunately I run into a critical bug in my system:

## Checklist

- [x] I have included information about relevant versions
- [x] I have verified that the issue persists when using the `master` branch of Faust.

## Steps to reproduce

1. fastapi_app.py - FastAPI app started with uvicorn.
2. faust_app.py - Faust app started with `python faust_app.py worker -l info`
3. both processes running in the same docker container.
4. Faust-streaming any version above 0.10.1 (checked 0.10.2, 0.10.3, 0.10.4)

Tell us what you did to cause something to happen.

## Expected behavior

I can import a topic created in faust_app.py into fastapi_app.py and use `await my_topic.send(value={...})` inside of FastAPI async endpoints (or use an async function that does the same). Everything works OK in faust-streaming version 0.10.1 and below.

## Actual behavior

`AssertionError: Please create objects with the same loop as running with`
or
`RuntimeError: Task attached to a different loop`
looks like the 1st attempt will result in the 1st error, and further attempts result in the second error.

## Full traceback

in production k8s
```pytb
....
  File "/app/my_service/fastapi_app.py", line 70, in send_my_message_endpoint
    await send_my_message(my_message)
  File "/app/my_service/faust_app.py", line 71, in send_my_message
    await my_message_topic.send(value=my_message.json())
  File "/usr/local/lib/python3.11/site-packages/faust/topics.py", line 189, in send
    return await self._send_now(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/faust/channels.py", line 314, in _send_now
    return await self.publish_message(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/faust/topics.py", line 429, in publish_message
    await producer.send(
  File "/usr/local/lib/python3.11/site-packages/faust/transport/drivers/aiokafka.py", line 1360, in send
    await transaction_producer.send(
  File "/usr/local/lib/python3.11/site-packages/aiokafka/producer/producer.py", line 442, in send
    await self.client._wait_on_metadata(topic)
  File "/usr/local/lib/python3.11/site-packages/aiokafka/client.py", line 639, in _wait_on_metadata
    await self.force_metadata_update()
RuntimeError: Task <Task pending name='Task-50' coro=<RequestResponseCycle.run_asgi() running at /usr/local/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py:407> cb=[set.discard()]> got Future <Future pending cb=[shield.<locals>._outer_done_callback() at /usr/local/lib/python3.11/asyncio/tasks.py:898]> attached to a different loop
```
in local docker-compose
```pytb
ERROR:    Exception in ASGI application
  File "/usr/local/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/fastapi/applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.11/site-packages/prometheus_fastapi_instrumentator/middleware.py", line 103, in __call__
    raise exc
  File "/usr/local/lib/python3.11/site-packages/prometheus_fastapi_instrumentator/middleware.py", line 101, in __call__
    await self.app(scope, receive, send_wrapper)
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/cors.py", line 92, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/cors.py", line 147, in simple_response
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/usr/local/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/routing.py", line 706, in __call__
    await route.handle(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/routing.py", line 66, in app
    response = await func(request)
               ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/fastapi/routing.py", line 235, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/fastapi/routing.py", line 161, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/my_service/fastapi_app.py", line 70, in send_my_message_endpoint
    await send_my_message(my_message)
  File "/app/my_service/faust_app.py", line 71, in send_my_message
    await my_message_topic.send(value=my_message.json())
  File "/usr/local/lib/python3.11/site-packages/faust/topics.py", line 189, in send
    return await self._send_now(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/faust/channels.py", line 314, in _send_now
    return await self.publish_message(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/faust/topics.py", line 407, in publish_message
    producer = await self._get_producer()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/faust/topics.py", line 384, in _get_producer
    return await self.app.maybe_start_producer()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/mode/utils/futures.py", line 58, in __call__
    result = await self.fun(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/faust/app/base.py", line 1569, in maybe_start_producer
    await producer.maybe_start()
  File "/usr/local/lib/python3.11/site-packages/mode/services.py", line 859, in maybe_start
    await self.start()
  File "/usr/local/lib/python3.11/site-packages/mode/services.py", line 800, in start
    await self._default_start()
  File "/usr/local/lib/python3.11/site-packages/mode/services.py", line 807, in _default_start
    await self._actually_start()
  File "/usr/local/lib/python3.11/site-packages/mode/services.py", line 824, in _actually_start
    await self.on_start()
  File "/usr/local/lib/python3.11/site-packages/faust/transport/drivers/aiokafka.py", line 1300, in on_start
    await producer.start()
  File "/usr/local/lib/python3.11/site-packages/aiokafka/producer/producer.py", line 303, in start
    assert self._loop is get_running_loop(), (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Please create objects with the same loop as running with
```
# Versions

* Python version 3.11.2
* Faust version 0.10.2, 0.10.3, 0.10.4
* Operating system Win11 + WSL (Debian container `python:3.11-slim`)
* Kafka version - `image: confluentinc/cp-kafka:7.0.0`

wbarnha:
## Description

Three related pieces of work on the web and observability layers:

1. Let an application replace the legacy `faust.web`/aiohttp server with any framework service, with a ready-made ASGI integration.
2. Get a single distributed trace across the HTTP request, Kafka, and the agent that consumes the message.
3. Control each built-in web endpoint individually, and expose worker performance metrics as JSON.

Fixes #448
Fixes #435
Fixes #322

Reviewable commit by commit, in order.

---

### 1. The event-loop fix

This is the bug behind all three issues, and it is not really a FastAPI bug at all.

Faust apps are declared at module scope, but under uvicorn the app is *started* from a loop created later by `asyncio.run()`. Several call sites resolved `app.loop` at **declaration** time, pinning the App to whichever loop `mode`'s `get_event_loop()` found at import — a loop that is never run. Everything built during startup then inherited that dead loop:

```
AssertionError: Please create objects with the same loop as running with
RuntimeError: Task ... got Future ... attached to a different loop
```

`faust worker` survived by accident: `faust/cli/base.py` reused the same policy loop `mode` had already grabbed, so the pinned loop happened to be the right one. `asyncio.run()` creates a different one, so the pin became fatal.

The fix is to stop resolving the loop eagerly and let `mode.Service` late-bind it inside `_default_start()` — i.e. in the loop that actually runs the app:

- `Agent`, `Collection`, `Web` and livecheck `Case` no longer pass `loop=app.loop` to `Service.__init__`.
- `App._new_transport()` / `_new_producer_transport()` no longer pass `loop=`, and `Transport.loop` resolves lazily (still settable; an explicit `loop=` still wins).
- `Transport.create_conductor()` no longer passes `loop=`.
- `App.tables` no longer passes `loop=` to the table manager.

**Worth a reviewer's eye:** the transport chain (`@app.agent` → `app.topics` → `_new_conductor()` → `app.transport`) is the load-bearing part. Fixing only the four `Service.__init__` sites leaves `app._loop` pinned — found by measuring, not reading.

A consequence caught by CI: nothing calls `set_event_loop()` before the CLI runs any more, and Python 3.14 removed the get-or-create fallback from `get_event_loop_policy().get_event_loop()`. `faust/cli/base.py` now routes through `mode.utils.loops.get_event_loop()`, which keeps get-or-create across versions. Only the 3.14 legs caught this; 3.10–3.13 still auto-create.

### 2. Pluggable web frameworks and `faust.contrib.asgi`

`App.web_server()` is the framework-neutral integration point. It accepts any `mode.Service` class or instance and makes it the worker's sole web server. A custom server:

- suppresses construction and startup of the legacy `faust.web`/aiohttp stack and its cache;
- starts after table recovery;
- shares the worker event loop;
- obeys `web_enabled` and `--without-web`;
- supplies worker banner URL and driver information without constructing aiohttp.

ASGI applications use:

```python
from faust.contrib.asgi import serve_asgi

api = FastAPI()
serve_asgi(faust_app, api)
```

The worker-driven path works with FastAPI, Starlette, Litestar, Quart, Django ASGI, or any other ASGI callable. The callable stays opaque to Faust, and uvicorn is imported lazily. Non-ASGI frameworks can provide their own `mode.Service`:

```python
@app.web_server
class MyWebServer(mode.Service):
    ...
```

When the ASGI application drives, frameworks with a lifespan hook can use `faust_lifespan()`:

```python
api = FastAPI(lifespan=faust_lifespan(faust_app))
```

Django exposes a standard ASGI callable but no constructor-level lifespan hook. `FaustLifespanMiddleware` supplies the standard ASGI startup and shutdown protocol without importing Django:

```python
from django.core.asgi import get_asgi_application
from faust.contrib.asgi import FaustLifespanMiddleware

django_application = get_asgi_application()
application = FaustLifespanMiddleware(django_application, faust_app)
```

The middleware owns only the lifespan scope. HTTP and WebSocket scopes pass to Django unchanged. The updated `examples/django/proj/asgi.py` demonstrates this path and removes the old Eventlet requirement. The example also uses current Django URL and translation APIs.

Host and port remain late-bound from `web_bind` and `web_port` when Faust drives through `serve_asgi()`, so CLI overrides remain effective. `faust.contrib.fastapi` remains as a compatibility alias.

### 3. OpenTelemetry

`opentelemetry-instrumentation-aiokafka` already wraps `AIOKafkaProducer.send` and `AIOKafkaConsumer.getmany`, so most of a trace already works. The hop nobody outside Faust can supply is `receive` → *process*: the consumer runs in its own thread (`ConsumerThread`) and `contextvars` never cross threads, so the `receive` span is opened *and closed* on a thread the agent never runs on. The result is an orphaned span and an unparented agent, which reads worse than no instrumentation.

`OpenTelemetrySensor` closes it — extracting trace context from the message headers and opening a `{topic} process` span that stays current while the stream processes the event:

```python
from faust.contrib.opentelemetry import setup_opentelemetry
setup_opentelemetry(app)
```

FastAPI apps are instrumented automatically, but only when OpenTelemetry is installed **and** a real `TracerProvider` is configured — until an SDK is configured the API is a no-op, so this never enables telemetry nobody asked for.

- Depends on `opentelemetry-api` only, and never calls `set_tracer_provider()` — configuring the SDK is the application's job.
- Reads trace context from headers, never writes it. The aiokafka instrumentation already injects on produce and its setter appends unconditionally, so a second injector would put two `traceparent` headers on the wire.
- Warns when the opentracing `TracingSensor` is registered, since that one *does* inject.
- Every OpenTelemetry call is wrapped — telemetry must never break message processing.

### 4. Per-endpoint web feature flags

Which built-in endpoints get served was decided entirely by `debug`: it enabled the statistics and graph endpoints together or neither, and `/router` and `/table` were mounted unconditionally with **no way to turn them off** — even though `/table` serves table *data* over HTTP and `/router` publishes the URL of every other worker.

```
web_stats_enabled     default_alias="debug"   /  and  /assignment/
web_graph_enabled     default_alias="debug"   /graph
web_router_enabled    default=True            /router
web_tables_enabled    default=True            /table
web_metrics_enabled   default=False           /performance/
```

`default_alias="debug"` resolves to `conf.debug` at read time when unset, so an app setting none of these serves exactly what it served before. `/router` and `/table` keep defaulting to on because disabling `/router` breaks `@table_route` across nodes — they are simply switchable now.

Selection moved into `Web._enabled_blueprints()`, driven by a `blueprint_flags` mapping. Blueprints with no entry are always enabled, so anything a subclass or user adds is unaffected.

### 5. `/performance/` metrics endpoint

Off by default, no new dependency, independent of `debug`. Not a flat `Monitor.asdict()` dump — two things are computed that monitor does not provide:

- **Consumer lag**, from log end offsets minus offsets actually read. Monitor tracks both but never subtracts them. Partitions with an unknown read offset are skipped rather than reported as fully lagged, which would spike alerts on every restart.
- **Latency percentiles**. Monitor keeps raw deques of thousands of samples. Nearest-rank rather than `statistics.quantiles`, which raises on fewer than two samples — a metrics endpoint should not fail on a worker that just started.

The payload builder now lives at `faust.sensors.metrics.performance_metrics()`, so custom frameworks can expose the same data without importing `faust.web`. The aiohttp endpoint is a thin adapter over that framework-neutral function.

## New extras

```
pip install "faust-streaming[asgi]"             # uvicorn for any ASGI application
pip install "faust-streaming[fastapi]"          # ASGI integration + FastAPI
pip install "faust-streaming[opentelemetry]"    # api + fastapi/aiokafka instrumentation
```

`requirements/test.txt` gains only `opentelemetry-api`, `-sdk` and `uvicorn` — all pure Python — so the new code is covered in CI without pulling `fastapi`/`pydantic-core` into all 15 matrix legs. FastAPI-dependent tests use `importorskip`; the OpenTelemetry instrumentation paths are covered by stubbing the module through `sys.modules`.

## Backwards compatibility

The existing aiohttp server remains the default when no custom server is registered, so existing applications keep their routes and startup behavior. `install_requires` is unchanged. `faust.contrib.fastapi` remains import-compatible with the new `faust.contrib.asgi` module.

Registering a custom server is intentionally exclusive: Faust does not construct or start the legacy `faust.web` server or its cache alongside it. Duplicate custom-server registration raises `ImproperlyConfigured` instead of silently choosing one implementation.

The earlier loop-late-binding audit still applies to the default path. The packaging fix also continues to exclude `examples.*` so example packages are not shipped in site-packages.

## Testing

2390 passed, 79 skipped across the full unit and functional suites locally. `mypy -p faust` reports no issues in 173 source files; `isort --jobs 1`, `black`, `flake8`, and `git diff --check` are clean. The non-strict Sphinx HTML build completes; strict mode is currently blocked by the pre-existing dead Redis intersphinx inventory URL.

New tests bind no socket and start no uvicorn, so they stay fast and keep the autouse lingering-thread/task guards happy. Full repository collection in this environment stops in the unrelated consistency checker because aiokafka 0.14 no longer exports `GroupCoordinatorRequest_v0`:

- `tests/functional/test_asgi_cohost.py` reproduces the import-time-vs-running-loop split without a broker, by declaring the app in a *synchronous* fixture and asserting from `async` tests. Verified these fail when the old code is restored.
- `tests/unit/contrib/test_asgi.py`, `tests/unit/app/test_base.py`, and `tests/unit/app/test_service.py` cover primary-server registration, recovery ordering, disabled-web behavior, late CLI settings, compatibility imports, suppression of legacy aiohttp components, ASGI lifespan protocol messages, failure signaling, and unchanged HTTP delegation.
- The Django example imports against Django 6.1 as a real `ASGIHandler`, and `django check` reports no issues.
- `tests/unit/sensors/test_metrics.py` covers the framework-neutral performance payload builder.
- `tests/unit/contrib/test_opentelemetry.py` asserts against a real `InMemorySpanExporter` that the process span continues the trace from a `traceparent` header and that nested spans parent to it.
- `tests/unit/web/test_blueprint_flags.py` covers the flag matrix and the back-compat cases.
- `tests/functional/web/test_metrics.py` drives `/performance/` through the existing aiohttp test client.

### Verified against a real broker

Both directions were run end to end against Kafka 4.0 (KRaft, single node):

```
faust worker + serve_asgi
  POST -> await topic.send() on the worker's loop -> 200
  agent in the same process consumed the message
  real SIGINT to the process group -> exit 0, 1.4s clean shutdown

uvicorn + faust_lifespan   (the literal #448 scenario)
  POST -> await topic.send() -> 200, zero loop errors
  agent consumed it, including the backlog from a previous run
```

### Signal handling

Checked against real SIGINTs — a `mode.Worker` hosting `AsgiService` with uvicorn serving on a real port, signalled at the process group like a terminal Ctrl-C, single and repeated, with and without the override. All four combinations exit 0 and complete a slow service's drain.

That measurement corrected the code comment: `mode.Worker` registers via `loop.add_signal_handler()`, which is delivered through asyncio's wakeup fd, so uvicorn's `signal.signal()` does not displace it. The override is kept to avoid a second concurrent shutdown path, and the docstring now says that rather than something untrue.

## Unrelated issues surfaced, not fixed here

- `extra/tools/render_configuration_reference.py` iterates `related_cli_options` string values character by character, so `make configref` rewrites `:option:`faust --debug`` as `:option:`faust -`, :option:`faust d`, ...` and drops hand-written prose. The new settings are hand-added to `settingref.txt` in the generator's format instead.
- `make configcheck` fails on master because `web_application_options` was never added to `settingref.txt`. Documented here so the symmetric difference is empty, but `configcheck` is a tox environment CI never invokes, which is why it drifted.
- The PyPy leg failed on a hypothesis `FlakyFailure` in the partition-assignor property test (5123ms against a 4000ms deadline on one run, 3455ms on the next). Pre-existing and unrelated; the deadline is raised on PyPy only, so CPython keeps the tighter bound.
- An agent will not consume from a topic that does not exist when the worker starts — messages are written but unprocessed until a restart. Confirmed to be general Faust behaviour (`faust worker` reproduces it identically), so it is documented in the userguide caveats rather than worked around.


wbarnha:
@copilot resolve the merge conflicts in this pull request

PR #759, opened 2026-08-06
Fix two crashes in the aiokafka threaded producer

wbarnha:
Both bugs were surfaced by the type checker in #758 and left marked `XXX` there, because fixing them changes runtime behaviour and that PR was annotation-only. This PR fixes them, each with a regression test.

> #758 has since been squash-merged, so this branch has been rebased onto `master` and now contains only the two fixes and their tests.

## 1. `_shutdown_thread` raised TypeError on every producer-thread shutdown

`ThreadedProducer._shutdown_thread` was a plain `def` overriding `mode.threads.ServiceThread._shutdown_thread`, which is `async def` and is awaited by `_serve()`. The fix makes it async and awaits `super()._shutdown_thread()`, preserving mode's teardown and the existing once-only guard.

## 2. `publish_message(wait=True)` could never succeed

The waiting branch called `Topic._on_published` without its required send-future argument. Real topics therefore raised `TypeError`.

The fix follows Faust's channel abstraction:

- `ChannelT` declares `_finalize_message`.
- The threaded producer reports send completion, then delegates result and callback handling to `channel._finalize_message`.
- This preserves channel-owned completion semantics and supports async callbacks.
- The non-waiting branch remains unchanged.

### Why the existing test missed it

The existing test used a bare `Mock()` channel, which accepts the malformed call. The regression test uses a real topic and verifies the future result, callback, and sensor completion.

## Tests

Added regression coverage for the real-topic wait path and all shutdown paths.


wbarnha:
`codecov/project` is red at **-0.01%** (96.05% → 96.04%). Everything else on `697a379` is green, including the required `✅ Ensure the required checks passing` gate, all 15 CPython legs, lint (`mypy -p faust`), CodeQL and both Kafka integration legs.

I'm not pushing a change for it, because there is no uncovered line to cover.

Codecov's own patch check agrees: *"All modified and coverable lines are covered by tests."* The project delta comes from its summary reporting `Lines +3, Hits +2, Misses +1` — but that extra miss is not reproducible. Diffing full-package coverage locally between `master` (`0866b77`) and this head, across all 109 measured files:

```
faust/transport/drivers/aiokafka.py:  master = 699 stmts / 49 miss
                                       #759  = 702 stmts / 49 miss
```

That is the *only* file whose numbers move anywhere in the package — 3 new statements, all 3 covered, misses flat. The missing lines are also the same set on both sides, just shifted by two line numbers.

So the -0.01% is a single line out of ~11,084 appearing as uncovered in the merged 15-leg report while every individual leg covers it. `codecov/project` runs on `target: auto` with `threshold: 0%`, so a one-line difference is enough to fail it.

Worth noting this is structural rather than specific to this PR: the repo has no `codecov.yml` at all, so codecov runs on defaults. A small config would stop both this and the transient red that every PR here shows for the first few minutes of a run (codecov scores partial uploads as legs finish, e.g. 5 of 16 → "32.50% (-63.56%)", then corrects itself):

```yaml
codecov:
  notify:
    after_n_builds: 15   # 5 pythons x 2 cython (aiokafka) + 5 confluent.
                         # Not 16 - the PyPy leg is continue-on-error and
                         # uploads conditionally, so waiting on it can hang.
coverage:
  status:
    project:
      default:
        threshold: 0.5%  # tolerate cross-leg variance
```

Happy to open that as a separate PR if wanted — it doesn't belong in this one.

---

PR #773, opened 2026-08-10
Support free-threaded CPython (PEP 703) on 3.13t and 3.14t

wbarnha:
A free-threaded interpreter re-enables the GIL for the whole process the
moment it imports an extension module that has not declared
`Py_mod_gil = Py_MOD_GIL_NOT_USED`, and reports it only through a
RuntimeWarning.  All three of faust's Cython extensions were in that
state, so importing faust on 3.13t/3.14t silently turned free-threading
off:

    RuntimeWarning: The global interpreter lock (GIL) has been enabled to
    load module 'faust._cython.windows', which has not declared that it
    can run safely without the GIL.

Set `freethreading_compatible=True` in the three .pyx files, which is
what makes Cython emit the slot.  The modules qualify: windows.pyx holds
cdef doubles written once in __init__ and read-only after, and
streams.pyx / conductor.pyx hold per-instance Python references with all
shared state in ordinary Python containers.

Two things made the loss easy to reintroduce invisibly, so both are
pinned down:

  * The directive only exists in Cython 3.1+, and older Cython ignores
    unknown directives rather than failing -- a 3.0 build would emit no
    declaration and no diagnostic.  Add a `cython>=3.1` floor for 3.13+
    in build-system.requires, and pin cibuildwheel's before-build the
    same way.

  * Nothing fails when the declaration is missing.  Add
    tests/unit/test_free_threading.py, which imports each extension in a
    subprocess and asserts the GIL is still off.  The subprocess drops
    PYTHON_GIL from its environment, or the CI job's PYTHON_GIL=0 would
    make the assertion vacuous.  It skips on a GIL interpreter.

With that, the full unit + functional suite passes on both 3.13t and
3.14t against the compiled extensions with the GIL genuinely disabled
(2211 passed), so drop the `cp31?t-*` cibuildwheel skip the previous
comment said to drop "once faust is verified free-threading-safe", and
add `enable = ["cpython-freethreading"]` so cp313t is built alongside
cp314t.

The new `free-threaded` job covers both interpreters and gates merges,
since it is what verifies the wheels being published.  Two things it
does differently from the other test jobs, both necessary:

  * It installs requirements/freethreading.txt, not test.txt.  Parts of
    test.txt cannot be built on a free-threaded interpreter at all --
    twine pulls in cffi, which refuses to build on 3.13t, and hypothesis
    6.130+ ships a PyO3 extension that does not support 3.13t either.
    The new file documents every omission and pin.

  * It builds the extensions in place.  pytest runs from the repo root,
    so `import faust` resolves to the source tree, and the accelerated
    implementations are imported behind `try: ... except ImportError`.
    Without a .so next to the .pyx the fallback engages silently and the
    run exercises pure Python regardless of USE_CYTHON -- which is also
    true of the existing USE_CYTHON=true matrix legs.

Document the above in docs/developerguide/free_threading.rst, along with
two findings that are not fixed here: aiokafka's extensions have not made
the declaration either, so a real worker gets the GIL back when the
transport driver loads; and Message.ack/decref is a non-atomic
read-modify-write that loses final acks under real parallelism.  The
latter is not reachable from faust's own code, which acks from the event
loop, but is reachable via the public Event.ack.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01K8qT5E3rnSXvw7ibNLXrVr

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -138,6 +138,66 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+  test-freethreading:
+    name: 'Python ${{ matrix.python-version }} (free-threaded)'
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        # The two interpreters `[tool.cibuildwheel]` publishes free-threaded
+        # wheels for.  Keep the two lists in step: a version we ship a wheel
+        # for is a version this job has to cover.
+        python-version: ['3.13t', '3.14t']
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+          cache-dependency-path: |
+            requirements/*.txt
+            requirements/extras/*.txt
+      - name: Install dependencies
+        # Not requirements/test.txt: parts of it cannot be built on a
+        # free-threaded interpreter at all (twine -> cffi, and hypothesis'
+        # PyO3 extension on 3.13t).  freethreading.txt is that list minus the
+        # ones that fail, and documents each omission.
+        run: |
+          pip install -r requirements/freethreading.txt
+          pip install 'Cython>=3.1' setuptools setuptools_scm
+          # Editable, unlike the other jobs' `pip install .`.  pytest runs from
+          # the repo root, so `import faust` resolves to the source tree either
+          # way -- but the suite also needs the distribution *metadata* to
+          # exist, because `faust/__init__.py` does
+          # `version("faust-streaming")` at import time.  An editable install
+          # registers that metadata against the tree the tests actually import,
+          # instead of a second copy in site-packages that nothing loads.
+          USE_CYTHON=1 pip install -e . --no-build-isolation
+      - name: Build the Cython extensions in place
+        # The extensions have to sit next to the .pyx files or they are never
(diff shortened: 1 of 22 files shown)


wbarnha:
Closing: this content is already on `master`. It merged via #762 as e9f102f, which carried every commit from this branch.

Verified per file rather than by commit, since the squash makes commit-level comparison unreliable — `git diff master <branch>` is empty for all of them:

```
faust/types/tuples.py                    same as master   (ack_lock)
faust/transport/consumer.py              same as master
faust/_cython/streams.pyx                same as master
tests/unit/test_ack_concurrency.py       same as master
tests/unit/test_free_threading.py        same as master
requirements/build.txt                   same as master
docs/includes/settingref.txt             same as master
docs/developerguide/free_threading.rst   same as master
```

The only tree difference runs the other way: `master` additionally has #746 and #753, which this branch predates.

Worth noting for anyone reading the diff above — it shows ~2300 lines as new, which is misleading. #762 was squash-merged, so this branch's commits are not ancestors of `master`, and GitHub computes the diff from the old merge base (e2d801a) instead. Nothing here is unmerged.

Two follow-ups from #762 remain open and are **not** covered by this PR or by `master`:

1. `pytest-run-parallel` is in `requirements/test.txt` but never invoked — no `--parallel-threads`, no markers, no references. It implies concurrency coverage that does not exist. The ack stress tests added in #762 use explicit threads with a `Barrier`, which is the right tool for that race; the plugin runs a whole test function in N threads, which does not reproduce it. Either wire it into a real job or drop the dependency.
2. The `pypy3.11` leg failed once on #762 (1 failed / 2229 passed, 11m48s against a 15m timeout). It is `continue-on-error: true` and outside the merge gate, so it did not block, and the failing test was never identified.

---

